PR #2, opened 2022-07-19
Setup dependencies

johnwatson484:
https://eaflood.atlassian.net/browse/SFI-2192﻿

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "ffc-pay-statement-generator",
-  "version": "0.1.1",
+  "version": "0.2.0",
   "description": "FFC payment statement generator",
   "homepage": "https://github.com/DEFRA/ffc-pay-statement-generator",
   "main": "app/index.js",


test/unit/messaging/index.test.js
```diff
@@ -0,0 +1,13 @@
...
+jest.mock('../../../app/data')
+const messageService = require('../../../app/messaging')
+
+describe('messaging', () => {
+  afterAll(async () => {
+    await messageService.stop()
+  })
+
+  test('runs', async () => {
+    await messageService.start()
```
simonnineteen90, line 11: does this test need an `expect`?

simonnineteen90 (commented):
just one comment ona test, might be my lack of understanding!

test/unit/messaging/index.test.js
johnwatson484: It doesn't as there's nothing to validate.  

The test will therefore pass if the starting the service does not throw an error.

To be honest, I'm not really a fan of a test like this but given there is no actual behaviour in this application yet, there's not a lot to really test happens.

simonnineteen90 (approved):

PR #3, opened 2022-07-25
add statement generator db schema

simonnineteen90:
https://eaflood.atlassian.net/browse/SFI-2196


app/data/models/generation.js
```diff
@@ -0,0 +1,13 @@
+module.exports = (sequelize, DataTypes) => {
+  return sequelize.define('generations', {
```
johnwatson484, line 2: Should be singular here I think, so `generation` rather than `generations`.

simonnineteen90:
SonarCloud is unhappy about the lack of test coverage for new code but agreed with John that not required for these changes. Therefore please disregard warning.

changelog/db.changelog-1.1.xml
```diff
@@ -0,0 +1,19 @@
...
+  <changeSet author="Simon Dunn" id="3">
+    <tagDatabase tag="v1.1.0" />
+  </changeSet>
+  <changeSet author="Simon Dunn" id="4">
+    <createTable tableName="generations">
+      <column autoIncrement="true" name="generationId" type="INTEGER">
+        <constraints nullable="false" primaryKey="true" primaryKeyName="generationId_pkey" />
+      </column>
+      <column name="statementData" type="JSON" />
+      <column name="generated" type="TIMESTAMP" />
```
AbidemiAdio, line 15: Consider a more descriptive name like `dateGenerated` for the field, at first read, `generated`  sounds like `boolean` field to me. Please let me know if you are happy with the change, otherwise I am happy to approve.

simonnineteen90: Agree with this and updated as suggested

AbidemiAdio (approved):

PR #5, opened 2022-08-03
Add Part 2 generation

johnwatson484:
https://eaflood.atlassian.net/browse/SFI-2141


package.json
```diff
@@ -17,13 +17,15 @@
...
-    "@hapi/hapi": "20.2.2",
-    "applicationinsights": "2.3.3",
     "@azure/identity": "2.0.5",
     "@azure/storage-blob": "12.10.0",
+    "@hapi/hapi": "20.2.2",
+    "applicationinsights": "2.3.3",
     "ffc-messaging": "2.5.1",
     "joi": "17.6.0",
     "log-timestamp": "0.3.0",
+    "moment": "^2.29.4",
```
marctemp, line 27: remove caret

app/messaging/schemas/funding.js
```diff
@@ -0,0 +1,16 @@
...
+  rate: Joi.string().optional().allow(''),
+  annualValue: Joi.string().required(),
+  quarterlyValue: Joi.string().required(),
+  quarterlyReduction: Joi.string().required(),
+  quarterlyPayment: Joi.string().required(),
+  reductions: Joi.array().items(Joi.object({
+    reason: Joi.string().required(),
+    value: Joi.string().required()
+  })).optional()
+})
```
marctemp, line 16: required or optional?

test/unit/messaging/schemas/address.test.js
```diff
@@ -0,0 +1,84 @@
+const schema = require('../../../../app/messaging/schemas/address')
```
marctemp, line 1: for all of the schema, tests checking tht error is thrown and specifically Error is thrown need given (sommit tht I dnt think many of the validation in our other services do but they should, right? could be a tech debt ticket)

test/unit/messaging/validate-statement.test.js
```diff
@@ -0,0 +1,48 @@
+const validateStatement = require('../../../app/messaging/validate-statement')
+const mockStatement = require('../../mock-statement-data')
+
+describe('validate statement', () => {
+  test('does not throw on valid statement', async () => {
+    expect(() => validateStatement(mockStatement)).not.toThrow()
+  })
+
+  test('throws on undefined statement', async () => {
+    expect(() => validateStatement(undefined)).toThrow()
```
marctemp, line 10: tests on it throwing correct Error class

Dockerfile
```diff
@@ -1,4 +1,4 @@
-ARG PARENT_VERSION=1.2.19-node16.15.1
+ARG PARENT_VERSION=2.1.0-node16.16.0
```
marctemp, line 1: Sommit we need to create tickets to do in all repos

app/generator/content/part1.js
```diff
@@ -0,0 +1,7 @@
+const part1 = (_statement) => {
```
marctemp, line 1: do we need to pass statement in if we dnt use it?

app/generator/content/part3.js
```diff
@@ -0,0 +1,7 @@
+const part3 = (_statement) => {
```
marctemp, line 1: do we need to pass statement in if we dnt use it?

app/generator/content/part2.js
```diff
@@ -0,0 +1,120 @@
+const part2 = (statement) => {
+  const reductions = mapReductions(statement.funding)
+
+  return [
+    { text: 'Part 2. Calculation', style: 'header2' },
+    { text: `We calculated the total ${statement.scheme.shortName} payment amount by adding together payments for all the standards which are part of your agreement, as shown in this table.` },
```
marctemp, line 6: w/out causing you PTSD: is the assumption here tht 1 statement corresponds to 1 scheme?

app/generator/create-log.js
```diff
@@ -0,0 +1,7 @@
+const db = require('../data')
+
+const createLog = async (statementData, filename, dateGenerated) => {
```
marctemp, line 3: Idk if we have a semi-pseudo dev practice of calling db save downs as functions with "save" in the name rather than `create`

app/generator/index.js
```diff
@@ -0,0 +1,24 @@
+const PdfPrinter = require('pdfmake')
+const moment = require('moment')
+const fonts = require('./fonts')
+const styles = require('./styles')
+const generateContent = require('./content')
+const createLog = require('./create-log')
+const publish = require('./publish')
+const printer = new PdfPrinter(fonts)
+
+const generateStatement = async (statement) => {
```
marctemp, line 10: thinking this func should have `try-catch`

test/integration/generator.test.js
```diff
@@ -0,0 +1,58 @@
...
+    }
+    expect(fileList.filter(x => x === `${config.folder}/${FILE_NAME}`).length).toBe(1)
+  })
+
+  test('updates log', async () => {
+    await processStatementMessage(message, receiver)
+    const log = await db.generation.findOne({ where: { filename: `${FILE_NAME}` } })
+    expect(log).not.toBeNull()
+    expect(log.statementData).toStrictEqual(mockStatement)
+    expect(log.dateGenerated).toStrictEqual(new Date(2022, 7, 5, 15, 30, 10, 120))
```
marctemp, line 51: split into sep. test cases at least the null check and the content check

test/integration/generator.test.js
```diff
@@ -0,0 +1,58 @@
...
+      body: mockStatement
+    }
+  })
+
+  afterAll(async () => {
+    await db.sequelize.truncate({ cascade: true })
+    await db.sequelize.close()
+  })
+
+  test('publishes statement to storage', async () => {
```
marctemp, line 37: test names should be more clear and descriptive

test/integration/generator.test.js
```diff
@@ -0,0 +1,58 @@
+const { BlobServiceClient } = require('@azure/storage-blob')
```
marctemp, line 1: test name needs changed, should reflect the name of the file the call is being made from, i.e. `process-statement-message.test.js`

test/integration/generator.test.js
```diff
@@ -0,0 +1,58 @@
...
+
+  test('updates log', async () => {
+    await processStatementMessage(message, receiver)
+    const log = await db.generation.findOne({ where: { filename: `${FILE_NAME}` } })
+    expect(log).not.toBeNull()
+    expect(log.statementData).toStrictEqual(mockStatement)
+    expect(log.dateGenerated).toStrictEqual(new Date(2022, 7, 5, 15, 30, 10, 120))
+  })
+
+  test('completes message', async () => {
```
marctemp, line 54: need test cases for:

- `validateStatement` throws
- `generateStatement` throws
- `receiver.completeMessage(message)` throws
- throw with `err.category === VALIDATION`

test/unit/generator/content/summary.test.js
```diff
@@ -0,0 +1,14 @@
+const summary = require('../../../../app/generator/content/summary')
+const mockStatement = require('../../../mock-statement-data')
+
+describe('generate summary', () => {
+  test('includes header', () => {
```
marctemp, line 5: when `scheme.name` is `Sust...` and `scheme.year` is `2022`

could have another test where these are changed

test/unit/generator/content/part2.test.js
```diff
@@ -0,0 +1,101 @@
...
+    const result = part2(mockStatement)
+    expect(result[1].text).toBe(`We calculated the total ${mockStatement.scheme.shortName} payment amount by adding together payments for all the standards which are part of your agreement, as shown in this table.`)
+  })
+
+  test('includes all table columns', () => {
+    const result = part2(mockStatement)
+    expect(result[2].table.body.length).toBe(6)
+  })
+
+  test('includes all table columns', () => {
```
marctemp, line 20: test names aren't descriptive enough, these have the same name but 2 different results

test/unit/generator/content/part2.test.js
```diff
@@ -0,0 +1,101 @@
...
+  })
+
+  test('includes help text introduction', () => {
+    const result = part2(mockStatement)
+    expect(result[3].text[0]).toBe('You can check which land parcels are part of each standard in the Rural Payments service at ')
+  })
+
+  test('includes help text link', () => {
+    const result = part2(mockStatement)
+    expect(result[3].text[1].link).toBe('https://www.gov.uk/claim-rural-payments')
```
marctemp, line 82: shows as new line but not in VSCode?

test/unit/generator/content/part2.test.js
```diff
@@ -0,0 +1,101 @@
...
+    expect(result[3].text[1].style).toBe('link')
+  })
+
+  test('includes help text full stop', () => {
+    const result = part2(mockStatement)
+    expect(result[3].text[2]).toBe('.')
+  })
+
+  test('includes reductions header if reductions', () => {
```
marctemp, line 91: test case when there aren't any reductions

test/unit/generator/create-filename.test.js
```diff
@@ -0,0 +1,52 @@
...
+describe('create filename', () => {
+  beforeEach(() => {
+    mockStatement = require('../../mock-statement-data')
+    jest.useFakeTimers().setSystemTime(new Date(2022, 7, 5, 15, 30, 10, 120))
+    timestamp = moment(new Date()).format('YYYYMMDDHHmmssSS')
+  })
+
+  test('writes full filename', () => {
+    const result = getFilename(mockStatement, timestamp)
+    expect(result).toBe('FFC_PaymentStatement_SFI_2022_1234567890_2022080515301012.pdf')
```
marctemp, line 17: again, shows new line but not in VSCode

marctemp: also, could you use `PREFIX` here?

test/unit/messaging/process-statement-message.test.js
```diff
@@ -1,23 +1,59 @@
...
+    }
+    mockValidation.mockImplementation(() => {
+      const err = new Error('Validation error')
+      err.category = VALIDATION
+      throw err
+    })
+    await processStatementMessage(message, receiver)
+    expect(receiver.completeMessage).not.toHaveBeenCalledWith(message)
+    expect(receiver.deadLetterMessage).toHaveBeenCalledWith(message)
+  })
```
marctemp, line 58: need test cases for:

- does it call `validateStatement`?
- how many times does it call `validateStatement`?
- what does it call `validateStatement` with?
- what if `validateStatement` throws?
- does it call `generateStatement`?
- how many times does it call `generateStatement`?
- what does it call `generateStatement` with?
- what if `generateStatement` throws?
- how many times does it call `receiver.completeMessage`?
- what if `receiver.completeMessage` throws?
- how many times does it call `receiver.deadLetterMessage`?
- what if `receiver.deadLetterMessage` throws?

test/unit/messaging/validate-statement.test.js
```diff
@@ -0,0 +1,48 @@
+const validateStatement = require('../../../app/messaging/validate-statement')
+const mockStatement = require('../../mock-statement-data')
+
+describe('validate statement', () => {
```
marctemp, line 4: do you not need to mock `schema` set it to return just `mockStatement` and then have tests around when it throws?

docs/asyncapi.yaml
```diff
@@ -0,0 +1,205 @@
...
+          description: The second line of the address
+          example: 'The Area'
+        line3:
+          type: string
+          description: The third line of the address
+          example: 'The Town'
+        line4:
+          type: string
+          description: The fourth line of the address
+          example: 'The Region'
```
marctemp, line 84: standardise using single quotes or dbl

app/generator/publish.js
```diff
@@ -0,0 +1,25 @@
+const { getOutboundBlobClient } = require('../storage')
```
marctemp, line 1: missing tests

docs/asyncapi.yaml
```diff
@@ -0,0 +1,205 @@
+asyncapi: 2.4.0
```
marctemp, line 1: `, optional` for keys optional in message

app/generator/content/part2.js
```diff
@@ -0,0 +1,120 @@
...
+      body: [
+        [
+          { text: 'Standard', style: 'tableHeader' },
+          { text: 'Level', style: 'tableHeader' },
+          { text: 'Rate', style: 'tableHeader' },
+          { text: 'Land area (ha)', style: 'tableHeader' },
+          { text: 'Annual value', style: 'tableHeader' },
+          { text: 'Quarterly value', style: 'tableHeader' },
+          { text: 'Quarterly reduction', style: 'tableHeader' },
+          { text: 'Quarterly reduction', style: 'tableHeader' }
```
marctemp, line 57: should be:

- Quarterly reduction
- Quarterly payment

app/generator/content/part2.js
```diff
@@ -0,0 +1,120 @@
...
+
+const getReductionList = (reductions) => {
+  const list = []
+  reductions.forEach(x => {
+    list.push(`(${x.id}) ${x.reason}`)
+  })
+  return list
+}
+
+module.exports = part2
```
marctemp, line 120: got some discrepancies using the `mock-statement-data`

Screenshot 2022-08-08 at 15 31 56

Quarterly payment for improved grasslands should be: 105-50 = 45 but is 92.50 (which is 105-(50/4) as if the qtr calc hadn't been done)

also, the total for payments (inc'ing the erroneous 92.50) is 100 quid more than it should be: actually 242.15

app/messaging/schemas/payment.js
```diff
@@ -0,0 +1,10 @@
+const Joi = require('joi')
+
+module.exports = Joi.object({
+  invoiceNumber: Joi.string().required(),
+  reference: Joi.string().required(),
+  dueDate: Joi.string().required(),
+  settled: Joi.string().required(),
+  calculated: Joi.string().required(),
+  value: Joi.string().required()
+})
```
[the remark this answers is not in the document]
johnwatson484: Required.

package.json
johnwatson484: A quick google tells me I can run this and stop me having to check this every time!

`npm config set save-exact true`

app/messaging/schemas/funding.js
johnwatson484: The values are required within the array item.  The array itself is optional.

app/generator/content/part1.js
johnwatson484: We will in the next ticket.  It's more just a placeholder to help whoever picks that up understand the flow.

app/generator/content/part3.js
johnwatson484: As above. It will on the next ticket.

app/generator/content/part2.js
johnwatson484: I wish someone could answer this.  So I've gone with the approach that this service will largely be consistent and independent of scheme.  For the current content it's pretty straight forward such as what you see here.

Most of part 2 is pretty generic too in terms of mapping.  Perhaps the area column would disappear display N/A for say Vets.

Part 3 will be a bit more wider change as it links to scheme guidance so maybe we'd have `part3SFI` `part3Vets` and just switch them.

However, later in the year I expect that a lot of this content will change anyway once we start getting real feedback.  That will likely happen long before we add new schemes.

So this repo will cover all schemes.  How complicated the mapping of content will be will depend.

app/generator/create-log.js
johnwatson484: It's just a single insert sql call so would effectively be a single transaction.

I'll rename to save.

johnwatson484: Renamed

app/generator/index.js
johnwatson484: It's already in a try catch at a higher level for processing of the original message that will capture the error.  What would be the advantage of a lower level catch that just throws the error again?

app/messaging/schemas/funding.js
marctemp: sorry, I was alluding to adding `.optional()` to the end of line 16 as done in the other schemas

app/generator/index.js
marctemp: Truesy -- ignore first comment

test/integration/generator.test.js
johnwatson484: I've got those covered through unit tests in `test/unit/messaging/process-statement-message.test.js`.  I saw this test file as a way of proving the inputs and outputs of the applications work.  Specifics and edge cases I've covered with lower level edge cases.

I didn't want to repeat effectively the same test here but do you think it would be worth it?

test/unit/generator/content/summary.test.js
johnwatson484: I did originally write a test for each known scheme but found as I was checking with a variable in the expect string they were effectively the same test over and over.

We could hard code the values into the test, but I don't want this service interface to be that tightly coupled to it's provider if it doesn't need to be.  eg if in later statement datasets, a decision is made to rename `Sustainable Farming Incentive` to `Farming Money Scheme`, as long as the contract is met in the message this would just map whatever is provided to the correct place.

I've renamed the test to make it a bit more descriptive.

johnwatson484: Ha realise I was talking about a totally different test...  Ignore 90% of what I said above.

test/integration/generator.test.js
marctemp: Aaaahhh you're abso'ly right -- could argue tht the last test here is (will be) duplicated within the unit test

test/unit/generator/content/part2.test.js
johnwatson484: The one above is a copy of one below without the correct name.  Fixed.

docs/asyncapi.yaml
marctemp: actually, looking at yer golden standard `async` doc it doesn't use quotes at all: https://github.com/DEFRA/ffc-pay-enrichment/blob/main/docs/asyncapi.yaml

test/unit/messaging/schemas/address.test.js
johnwatson484: The schema validation itself doesn't throw an error.  It more just creates an error property and then it's up to the application to check that result and decide if it should throw an error.

As these files are scoped to the schema, then I wouldn't expect to check errors are thrown.  But maybe we could check the contents of the error property more than we are.  Might be worth digging into that property definition a bit more as wouldn't want to hard code tests text too specifically and end up resulting in failures if Joi changes internally.  But agree we could do more.  Let me do some reading.

test/unit/messaging/process-statement-message.test.js
johnwatson484: Added tests to cover the majority of these.

`deadLetterMessage` and `completeMessage` are a bit tricky as they're Service Bus functions.  So the actual behaviour of those failing is handled by the Azure SDKs.  They will send a message back to Service Bus to state to log the processing failure.  There's no real way for us to test that, particularly as we're mocking the SDK in the unit tests.

Dockerfile
johnwatson484: We should be regularly updating them as we work on them.  Just like updating packages to deal with vulns frequently, we should do the same with base images

johnwatson484: Challenge is, how do we do it regularly for inactive repos?

app/generator/index.js
```diff
@@ -0,0 +1,24 @@
+const PdfPrinter = require('pdfmake')
```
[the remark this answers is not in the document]
johnwatson484: I didn't add any here as all behaviour was already covered in higher level tests.

app/generator/publish.js
johnwatson484: Similar to other comment, all behaviour is covered in higher level tests.   It's also pretty difficult to setup a test at this level as the event emitters are awkward.  I felt it was a lot of effort to cover nothing new.

What do you think?

app/messaging/schemas/funding.js
johnwatson484: Got you! `funding` and `payments` are array items rather than child properties so don't need `required` or `optional`.  Within `statement` have specified there must be at least 1.

For `scheme` etc I could move required to be referenced in statement instead if you think that's clearer?

docs/asyncapi.yaml
johnwatson484: The async api spec only has a required flag for object schemas, https://www.asyncapi.com/docs/reference/specification/v2.0.0#schemaObject

Am I missing an optional?

docs/asyncapi.yaml
johnwatson484: I blame co-pilot haha

app/generator/content/part2.js
```diff
@@ -0,0 +1,120 @@
...
+        'You can check which land parcels are part of each standard in the Rural Payments service at ',
+        { text: 'www.gov.uk/claim-rural-payments', link: 'https://www.gov.uk/claim-rural-payments', style: 'link' },
+        '.'
+      ]
+    },
+    ...getReductions(reductions)
+  ]
+}
+
+const mapReductions = (funding) => {
```
marctemp, line 19: do you think this and the funcs below need unit tested?

appreciate you have unit tests for the `part2` func which calls out to these but it deals more with arranging how the outputs of these are displayed, rather than how they are created

test/unit/generator/create-filename.test.js
```diff
@@ -0,0 +1,52 @@
...
+
+  test('has PDF extension', () => {
+    const result = getFilename(mockStatement, timestamp)
+    expect(result.endsWith(EXTENSION)).toBeTruthy()
+  })
+
+  test('includes scheme short name', () => {
+    mockStatement.scheme.shortName = 'SFI'
+    const result = getFilename(mockStatement, timestamp)
+    expect(result).toContain('_SFI_')
```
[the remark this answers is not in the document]
marctemp: ^^ could you add similar tests for cases below too?

test/unit/messaging/process-statement-message.test.js
```diff
@@ -1,23 +1,88 @@
...
 
-  afterEach(() => {
-    jest.clearAllMocks()
+  test('completes message on success', async () => {
+    const message = {
+      body: mockStatement
+    }
+    await processStatementMessage(message, receiver)
+    expect(receiver.completeMessage).toHaveBeenCalledWith(message)
+    expect(receiver.completeMessage).toHaveBeenCalledTimes(1)
```
marctemp, line 31: split into 2 tests

test/unit/messaging/process-statement-message.test.js
```diff
@@ -1,23 +1,88 @@
...
+const mockValidation = jest.fn()
+jest.mock('../../../app/messaging/validate-statement', () => {
+  return mockValidation
+})
 const processStatementMessage = require('../../../app/messaging/process-statement-message')
+const mockStatement = require('../../mock-statement-data')
+const { VALIDATION } = require('../../../app/errors')
 let receiver
 
 describe('process statement message', () => {
```
marctemp, line 16: tests need splitting into single asserts

test/unit/messaging/process-statement-message.test.js
```diff
@@ -1,23 +1,88 @@
...
+  })
+
+  test('does not complete message on error', async () => {
+    const message = {
+      body: mockStatement
+    }
+    mockGenerator.mockImplementation(() => { throw new Error('Unable to generate') })
+    await processStatementMessage(message, receiver)
+    expect(receiver.completeMessage).not.toHaveBeenCalledWith(message)
+    expect(receiver.deadLetterMessage).not.toHaveBeenCalledWith(message)
```
marctemp, line 59: think this needs tweaking to: `expect(receiver.completeMessage).not.toHaveBeenCalled()`

the existing test would pass if both of these funcs were called w/ anything other than `message`, however we dnt want them to be called at all which the tweaked test handles

test/unit/messaging/validate-statement.test.js
marctemp: bumpin' in case missed

test/unit/messaging/schemas/statement.test.js
```diff
@@ -0,0 +1,78 @@
...
+    const result = schema.validate(mockStatement)
+    expect(result.error).toBeDefined()
+  })
+
+  test('validates fail if missing scheme', () => {
+    delete mockStatement.scheme
+    const result = schema.validate(mockStatement)
+    expect(result.error).toBeDefined()
+  })
+})
```
marctemp, line 78: need test for:

- email given
- email not given
- invalid email given

app/generator/content/part2.js
johnwatson484: Yes I think that's a fair comment.  I was considering breaking down part 2 into a more full module rather than a single file so this comment has persuaded me this is the right thing to do.  

Some of these low level functions have a decent degree of complexity so it does make sense to have more targeted tests against them to cover any edge cases.  For the table itself not so much (I've added some anyway) but for the others it would be extra assurance.  Particularly when we start mapping other schemes to the table.

test/unit/messaging/process-statement-message.test.js
johnwatson484: Yes true.  Fixed

app/generator/content/part2/get-table/map-total-row.js
```diff
@@ -0,0 +1,14 @@
+const mapTotalRow = (row) => {
```
marctemp, line 1: needs to handle if `row` is `undefined` -- used in `getTable` via  a `[].find` which returns `undefined` if it doesn't find anything that satisfies

app/generator/content/part2/get-table/map-funding-rows.js
```diff
@@ -0,0 +1,16 @@
+const mapReductionIds = require('./map-reduction-ids')
+
+const mapFundingRows = (funding, reductions) => {
```
marctemp, line 3: naming convention, array should be plural: `fundings`

app/generator/content/part2/get-table/index.js
```diff
@@ -0,0 +1,32 @@
+const mapFundingRows = require('./map-funding-rows')
+const mapTotalRow = require('./map-total-row')
+
+const getTable = (funding, reductions) => {
```
marctemp, line 4: naming convention, array should be plural: fundings

app/generator/content/part2/map-reductions.js
```diff
@@ -0,0 +1,20 @@
+const mapReductions = (funding) => {
```
marctemp, line 1: naming convention, array should be plural: fundings

test/unit/generator/content/part2/get-reductions.test.js
```diff
@@ -0,0 +1,30 @@
+const getReductions = require('../../../../../app/generator/content/part2/get-reductions')
+
+describe('get reductions', () => {
```
marctemp, line 3: test cases testing the `.style` returned

test/unit/generator/content/part2/get-table/get-table.test.js
```diff
@@ -0,0 +1,60 @@
+const getTable = require('../../../../../../app/generator/content/part2/get-table')
+let mockStatement
+let reductions
+
+describe('get table', () => {
+  beforeEach(() => {
+    mockStatement = JSON.parse(JSON.stringify(require('../../../../../mock-statement-data')))
+    reductions = [{ id: 1, reason: 'Late claim submission' }, { id: 2, reason: 'Over declaration reduction' }]
```
marctemp, line 8: pull out into own mock object sim. to `mock-statement-data`

used in multiple files and we can see the benefit:

- in here and `get-table/map-funding-rows.test.js` they resemble real-life examples
- but in `get-table/map-reduction-ids.test.js` they're placeholder
- not to mention making the tests familiar, predictable and orientated to the business process/output

test/unit/generator/content/part2/get-table/map-funding-rows.test.js
```diff
@@ -0,0 +1,60 @@
+const mapFundingRows = require('../../../../../../app/generator/content/part2/get-table/map-funding-rows')
+let mockStatement
+let reductions
+
+describe('map funding rows', () => {
```
marctemp, line 5: test cases for the styles given against each

test/unit/generator/content/part2/get-table/map-funding-rows.test.js
```diff
@@ -0,0 +1,60 @@
+const mapFundingRows = require('../../../../../../app/generator/content/part2/get-table/map-funding-rows')
```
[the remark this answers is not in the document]
marctemp: tests need added for:

- does it call `mapReductionIds`
- how many times does it call `mapReductionIds`
- what does it call `mapReductionIds` with
- what if `mapReductionIds` returns any of its __3__(?) edge cases?
- discuss: what if `mapReductionIds` thows

test/unit/generator/content/part2/get-table/map-reduction-ids.test.js
```diff
@@ -0,0 +1,42 @@
...
+  test('returns reduction id when one reduction', () => {
+    const result = mapReductionIds([{ reason: 'reason' }], reductions)
+    expect(result).toMatch('\n(1)')
+  })
+  test('returns reduction ids when multiple reductions', () => {
+    const result = mapReductionIds([{ reason: 'reason' }, { reason: 'reason2' }], reductions)
+    expect(result).toMatch('\n(1)\n(2)')
+  })
+
+  test('returns reduction ids when multiple reductions', () => {
```
marctemp, line 33: dup test from above

test/unit/generator/content/part2/get-table/map-reduction-ids.test.js
```diff
@@ -0,0 +1,42 @@
...
+    const result = mapReductionIds([{ reason: 'reason' }, { reason: 'reason2' }], reductions)
+    expect(result).toMatch('\n(1)\n(2)')
+  })
+
+  test('ignores unknown reductions', () => {
```
marctemp, line 38: seems a bit of a risk? 

happy to leave as-is and come back to if needs be

marctemp: can you create a test case where the unknown is in the middle of 2 knowns, just ensuring it doesn't mess up the formatting?

test/unit/generator/content/part2/get-table/map-total-row.test.js
```diff
@@ -0,0 +1,55 @@
+const mapTotalRow = require('../../../../../../app/generator/content/part2/get-table/map-total-row')
+let totalRow
+let reductions
+
+describe('map total row', () => {
```
marctemp, line 5: again, tests on styles and bold bool for these

test/unit/generator/content/part2/map-reductions.test.js
```diff
@@ -0,0 +1,50 @@
...
+    const result = mapReductions([{ reductions: [{ reason: 'reason' }] }])
+    expect(result[0].id).toBe(1)
+  })
+
+  test('maps reason to reduction if one reduction', () => {
+    const result = mapReductions([{ reductions: [{ reason: 'reason' }] }])
+    expect(result[0].reason).toBe('reason')
+  })
+
+  test('returns reductions if multiple reductions', () => {
```
marctemp, line 34: test for mapping only 1 id and 1 reason when multiple reductions but duplicated

marctemp: agains couple of scenarios for where those duplicates live:

- reductionA, reductionA
- reductionA, reductionB, reductionA

app/generator/content/part2/map-reductions.js
```diff
@@ -0,0 +1,20 @@
...
+  const reductions = []
+  let i = 1
+  funding.forEach(x => {
+    if (x.reductions) {
+      x.reductions.forEach(y => {
+        if (!reductions.find(z => z.reason === y.reason)) {
```
marctemp, line 7: suggestion: personally, I think `!==` inside reads easier rather than `===` and negating the whole thing

test/unit/generator/content/part2/part2.test.js
```diff
@@ -0,0 +1,109 @@
...
+    const result = part2(mockStatement)
+    expect(result[0].text).toBe('Part 2. Calculation')
+  })
+
+  test('includes introduction with scheme short name', () => {
+    const result = part2(mockStatement)
+    expect(result[1].text).toBe(`We calculated the total ${mockStatement.scheme.shortName} payment amount by adding together payments for all the standards which are part of your agreement, as shown in this table.`)
+  })
+
+  test('includes all table rows when five funding options and total row', () => {
```
marctemp, line 19: is it supposed to be, "_with_ five funding options.."?

test/unit/generator/content/part2/part2.test.js
```diff
@@ -0,0 +1,109 @@
+const part2 = require('../../../../../app/generator/content/part2')
```
marctemp, line 1: need to mock `getReductions`, `getTable` and `mapReductions` for unit test


think you can move this file into integration test

test/unit/generator/content/part2/part2.test.js
```diff
@@ -0,0 +1,109 @@
...
+let mockStatement
+
+describe('generate part 2', () => {
+  beforeEach(() => {
+    mockStatement = JSON.parse(JSON.stringify(require('../../../../mock-statement-data')))
+  })
+
+  test('includes header', () => {
```
marctemp, line 9: finally, tests  for style, etc

test/unit/generator/content/part2/part2.test.js
```diff
@@ -0,0 +1,109 @@
...
+  test('includes help text link', () => {
+    const result = part2(mockStatement)
+    expect(result[3].text[1].link).toBe('https://www.gov.uk/claim-rural-payments')
+    expect(result[3].text[1].style).toBe('link')
+  })
+
+  test('includes help text full stop', () => {
+    const result = part2(mockStatement)
+    expect(result[3].text[2]).toBe('.')
+  })
```
marctemp, line 88: genuinely I think my fav test there's ever been in these services -- shows yer attention to understanding EXACTLY how this service should look -- good job :)

test/unit/messaging/process-statement-message.test.js
```diff
@@ -1,23 +1,109 @@
...
+    const message = {
+      body: mockStatement
+    }
+    mockValidation.mockImplementation(() => {
+      const err = new Error('Validation error')
+      err.category = VALIDATION
+      throw err
+    })
+    await processStatementMessage(message, receiver)
+    expect(receiver.completeMessage).not.toHaveBeenCalledWith(message)
```
marctemp, line 93: missed, `.not.toHaveBeenCalled()`

marctemp: again, would suggest splitting

test/unit/messaging/process-statement-message.test.js
```diff
@@ -1,23 +1,109 @@
...
+  test('does not dead letter message on non-validation error', async () => {
+    const message = {
+      body: mockStatement
+    }
+    mockGenerator.mockImplementation(() => {
+      throw new Error('A generation error')
+    })
+    await processStatementMessage(message, receiver)
+    expect(receiver.completeMessage).not.toHaveBeenCalledWith(message)
+    expect(receiver.deadLetterMessage).not.toHaveBeenCalledWith(message)
```
marctemp, line 107: missed, `.not.toHaveBeenCalled()`

marctemp: again, would suggest splitting

app/generator/content/part2/map-reductions.js
johnwatson484: It wouldn't work with !== as then it would return as undefined for first item and then defined for second when no match.  Could use `some` rather than `find` if that's clearer?

test/unit/generator/content/part2/get-table/map-funding-rows.test.js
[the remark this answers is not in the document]
johnwatson484: Not sure I see the value in mocking the lower level function.  As long as we've covered all possible behaviour expected of the higher function.  How many sub functions it's composed of is less important.  Unless they result in specific behaviour like updating dependencies or logging messages say.

test/unit/generator/content/part2/get-table/map-reduction-ids.test.js
johnwatson484: Got too excited

test/unit/generator/content/part2/get-table/map-reduction-ids.test.js
johnwatson484: It's not actually possible for this to happen, but something told me to write this test.  Premonition?!

Added suggested test, all good

test/unit/generator/content/part2/part2.test.js
johnwatson484: I still think it's a unit test.  We've just decomposed the unit into smaller cleaner units.  It doesn't cross any application layers so don't see it as integration.

This one may in time be worth mocking the sub functions if the complexity to build the content grows, but I don't think it's necessary yet.

test/unit/generator/content/part2/part2.test.js
johnwatson484: I'm going to add this to my CV

test/unit/generator/content/part2/get-table/map-total-row.test.js
```diff
@@ -0,0 +1,125 @@
...
+    const result = mapTotalRow(totalRow, reductions)
+    expect(result.length).toBe(8)
+  })
+
+  test('returns empty array if undefined total row', () => {
+    const result = mapTotalRow(undefined, reductions)
+    expect(result.length).toBe(0)
+  })
+
+  test('includes funding option', () => {
```
marctemp, line 21: going to mention here but overlaps with lots of test files

does funding option (in this case) not havva style associated? idk the pdfprinter but does it not use the default style in-place when 1 isn't given? should we add tests for tht if so?

test/unit/generator/content/part2/map-reductions.test.js
```diff
@@ -0,0 +1,88 @@
...
+        { reason: 'Late claim submission' },
+        { reason: 'Over declaration reduction' },
+        { reason: 'Late claim submission' },
+        { reason: 'Late amendment reduction' }
+      ]
+    }])
+    expect(result[0].reason).toBe('Late claim submission')
+    expect(result[1].reason).toBe('Over declaration reduction')
+    expect(result[2].reason).toBe('Late amendment reduction')
+    expect(result.length).toBe(3)
```
marctemp, line 86: nice tests here -- could you split into single/more granular assertions:

- `length`
- each `id`
- each `reason`

same as you did for lines 34 - 49

test/unit/messaging/schemas/address.test.js
marctemp: Yup, agree on 1st point.

I _think_ what I was getting at was more around your `validate-statement` file and func which specifically runs the message against the schema, checks for any appended `.error` key and throws throw if so

Our other services do those steps fine but they usually do it w/in a larger func -- having this in its own means we can throw an error and catch it in those larger funcs.

I think my thinking was to adopt this way and implement across our other services, not sure?

test/unit/messaging/schemas/payment.test.js
```diff
@@ -0,0 +1,84 @@
+const schema = require('../../../../app/messaging/schemas/payment')
+let mockPayment
+
+describe('payment schema', () => {
+  beforeEach(() => {
+    mockPayment = JSON.parse(JSON.stringify(require('../../../mocks/statement-data').payments[0]))
+  })
+  test('validates success if all present', () => {
```
marctemp, line 8: whitespace

test/unit/generator/content/part2/map-reductions.test.js
johnwatson484: I've split by reasons and length as suggested.  I did try each category but  felt that it was harder for a reader to interpret the expected outcome to the input without knowing which other tests that fit the collection.

test/unit/generator/content/part2/get-table/map-total-row.test.js
johnwatson484: Done.  I've split the doc definition into it's own module and added some tests that cover those properties beyond the content.

test/unit/generator/content/part2/get-reductions.test.js
```diff
@@ -0,0 +1,36 @@
...
+  test('returns empty string in array if no reductions', () => {
+    const result = getReductions([])
+    expect(result.length).toBe(1)
+    expect(result[0]).toBe('')
+  })
+
+  test('includes reductions header if reductions', () => {
+    const reductions = [{ id: 1, reason: 'Late claim submission' }]
```
marctemp, line 11: can we use `mockReductions` here and throughout?

const reductions = mockReductions[0]

test/unit/generator/content/part2/map-reductions.test.js
```diff
@@ -0,0 +1,116 @@
...
+    expect(result.length).toBe(0)
+  })
+
+  test('returns empty array if no funding', () => {
+    const result = mapReductions([])
+    expect(result.length).toBe(0)
+  })
+
+  test('returns reductions if one reduction', () => {
+    const result = mapReductions([{ reductions: [{ reason: 'Late claim submission' }] }])
```
marctemp, line 20: Thoughts on using `mockReductions` here and throughout?

Maybe alludes to best having the reasons as `constants` in `app` and importing them into `mockReductions` and here, whatcha think?

test/integration/process-statement-message.test.js
```diff
@@ -0,0 +1,70 @@
...
+    expect(log.statementData).toStrictEqual(mockStatement)
+  })
+
+  test('saves log entry with generation date', async () => {
+    await processStatementMessage(message, receiver)
+    const log = await db.generation.findOne({ where: { filename: `${FILE_NAME}` } })
+    expect(log.dateGenerated).toStrictEqual(new Date(2022, 7, 5, 15, 30, 10, 120))
+  })
+
+  test('completes message', async () => {
```
marctemp, line 66: duplicate test in `unit/process-statement-message.test.js` -- think this can be deleted

test/integration/process-statement-message.test.js
```diff
@@ -0,0 +1,70 @@
...
+    await processStatementMessage(message, receiver)
+    expect(receiver.completeMessage).toHaveBeenCalled()
```
marctemp, line 68: no integration test possible for invalid message becuz we dnt check/connect to ServiceBus, i.e. check that cleared down Topic has 0 messages in both Queue and DLQ, then run function w/ good request (see Queue have 1 message, verify its the one we sent) or bad request (see DLQ [after a while due to retry] have 1 message, verify its the one we sent)

We dnt do this because we have no local Azure ServiceBus emulator, right?

test/unit/generator/content/part2/map-reductions.test.js
johnwatson484: The mock reductions object is mocking the output of the function this test class is testing.  So couldn't use it here.

I did originally use the `mockStatement` object for a few of the tests, but without adding more reductions to it to cover all of these test scenarios all it brought was inconsistency and difficulty understanding what the input to the test was.

test/integration/process-statement-message.test.js
johnwatson484: Yes that's the main reason we don't.  If had an emulator we would do exactly as you suggest, similar to what we do with Postgres when saving records.  Means our tests would give us much more confidence as they're using real protocols and integrations.

It is possible to do without emulation and we have done it in the demo service, but it involves pain.

The Jenkins pipeline will provision a full set of topics/subscriptions based on the contents of `provision.azure.yaml` unique to that build.  So every CI build does have a set of collision free queues with the intention they can be used for integration tests.  Once the build is complete those queues are automatically deleted.

But if you've got multiple tests writing messages to those queues, because of the async nature of messaging you can't necessarily guarantee those messages will be pulled off the queue in the right order.  So you need to think about how your tests wait for the right message for the specific test.  You also need to run test not in parallel (as we do anyway so we can managed connections consistently)  and have them setup to clear the queues before and after each test execution.  So you end up with more service bus processing code in your tests to do this as there isn't a nice `queue.clear()` command.

Ideally you'd have a unique set of queues per build per test, but that's a lot of work in Jenkins we haven't done to solve that problem.  There is a service bus sdk that supports creation and deletion of entities from Node.js so another solution would be for tests to manage their dependencies themselves.  But what if a test fails in CI, the clean up steps don't run and we end up with a treasure trove of orphan queues over time without further effort to make sure they're always cleaned up.

The other problem it gives us is local development outside of CI.  We have one set of queues per developer so they would share the same queues as local running of tests as they would for running the app.  Again the tests could empty the queues before they run each one, but what if I've spent ages loading my queue with messages for local dev.  I don't want the tests to purge all that.

Finally, we've been using service bus long enough that we've got high confidence that if we call `complete`, `deadLetter` , `abandon` etc then service bus will do as we ask.  So there is risk associated with not testing it, but the risk is so low it's just not worth all the pain to do it.

One positive I did notice though from Microsoft is this comment on the (long) open GitHub issue for local emulation.

image

https://github.com/Azure/azure-service-bus/issues/223

Just because it's on a backlog doesn't mean it will get done but it's the most positive sign I've seen in years!

test/unit/generator/content/part2/get-table/map-total-row.test.js
```diff
@@ -0,0 +1,125 @@
...
+  })
+
+  test('includes funding rate', () => {
+    const result = mapTotalRow(totalRow, reductions)
+    expect(result[2].text).toBe(totalRow.rate)
+  })
+
+  test('includes funding rate in bold', () => {
+    const result = mapTotalRow(totalRow, reductions)
+    expect(result[2].bold).toBeTruthy()
```
marctemp, line 48: `result[2]` has `.style`

test/unit/generator/content/part2/part2.test.js
```diff
@@ -0,0 +1,114 @@
...
+
+describe('generate part 2', () => {
+  beforeEach(() => {
+    mockStatement = JSON.parse(JSON.stringify(require('../../../../mocks/statement-data')))
+  })
+
+  test('includes header', () => {
+    const result = part2(mockStatement)
+    expect(result[0].text).toBe('Part 2. Calculation')
+    expect(result[0].style).toBe('header2')
```
marctemp, line 12: split into 2 (soz to be that guy but yer use to it now :{P)

test/unit/generator/content/part2/part2.test.js
```diff
@@ -0,0 +1,114 @@
...
+  })
+
+  test('includes all table rows with five funding options and total row', () => {
+    const result = part2(mockStatement)
+    expect(result[2].table.body.length).toBe(6)
+  })
+
+  test('includes all table columns', () => {
+    const result = part2(mockStatement)
+    expect(result[2].table.body[0].length).toBe(8)
```
marctemp, line 27: `result[2].style` to `table`

marctemp: test need added, this test is foine :D

marctemp (commented):
Last few changes/suggestions added

test/unit/generator/content/part2/part2.test.js
johnwatson484: Thought I'd got them all!  Darn

test/unit/generator/content/part2/get-table/map-total-row.test.js
```diff
@@ -48,6 +48,11 @@ describe('map total row', () => {
     expect(result[2].bold).toBeTruthy()
   })
 
+  test('includes funding rate style bold', () => {
```
marctemp, line 51: one final copy-pasta and this baby is ready to-go

marctemp (approved):
"Some of the best code I've ever read"
    - IGN

PR #6, opened 2022-08-06
Add part 3

johnwatson484:
https://eaflood.atlassian.net/browse/SFI-2234


app/generator/content/part1.js
```diff
@@ -1,7 +1,10 @@
 const part1 = (_statement) => {
-  return [
-    { text: 'Part 1. Payment summary', style: 'header2' }
-  ]
+  return {
+    stack: [
+      { text: 'Part 1. Payment summary', style: 'header2' }
+    ],
+    unbreakable: true
```
johnwatson484, line 6: Changing all parts beyond summary to have their content within `stacks`.  This allows us to easily ensure that each part stays on the same page with the `unbreakable` property.  Tests updated to accommodate the new structure and ensure that unbreakable is set on part 1, 2 and 3.

AbidemiAdio (approved):
I can see that work is in progress on part1.js in another branch, hence, this review is limited to part3.js and directly related files.

QED

PR #12, opened 2022-08-23
Support local development with publisher

johnwatson484:


app/config/storage.js
```diff
@@ -4,7 +4,7 @@ const Joi = require('joi')
 const schema = Joi.object({
   connectionStr: Joi.string().when('useConnectionStr', { is: true, then: Joi.required(), otherwise: Joi.allow('').optional() }),
   storageAccount: Joi.string().required(),
-  container: Joi.string().default('statement'),
+  container: Joi.string().default('statements'),
```
johnwatson484, line 7: Correcting name of container so consistent with higher environments and publisher repo

docker-compose.link.yaml
```diff
@@ -0,0 +1,5 @@
+version: '3.8'
+networks:
+  default:
+    external:
+      name: ffc-pay
```
johnwatson484, line 5: Required so containers are available across docker compose projects.

Same approach taken in payment service, SFI and demo etc

docker-compose.override.yaml
```diff
@@ -14,15 +14,15 @@ services:
...
+      AZURE_STORAGE_CONNECTION_STRING: DefaultEndpointsProtocol=http;AccountName=devstoreaccount1;AccountKey=Eby8vdM02xNOcqFlqUwJPLlmEtlCDXJ1OUzFT50uSRZ6IFsuFq2UVErCz4I6tq/K1SZFPTOtr/KBHBeksoGMGw==;BlobEndpoint=http://ffc-pay-statement-azurite:10000/devstoreaccount1;
 
   ffc-pay-statement-generator-postgres:
     ports:
       - "5451:5432"
     volumes:
       - postgres_data:/var/lib/postgresql/data
 
-  ffc-pay-statement-generator-azurite:
+  ffc-pay-statement-azurite:
```
johnwatson484, line 25: Change container to be consistent across both `publisher` and `generator` projects.  This means that both can target the same container endpoint.

scripts/start
```diff
@@ -4,6 +4,11 @@ if [ -z "$(docker network ls --filter name=^ffc-pay$ --format={{.Name}})" ]; the
   docker network create ffc-pay
 fi
 
+if [ -z "$(docker container ls --filter name=ffc-pay-statement-azurite  --format={{.ID}})" ]; then
+  echo "Azurite container already exists, skipping creation"
+  args="--scale ffc-pay-statement-azurite=0"
+fi
```
johnwatson484, line 10: This is the main part of this PR.  If there is already a container running named `ffc-pay-statement-storage` then don't create a conflicting container with the same name when starting this repo.  ie, scale the number of instances to 0.

The result of this will mean that this repo will just use the existing container rather than creating a conflicting one.  Therefore both services will be able to access the same files for local development.

Making the same changes in the publisher repo also means it doesn't matter which way around these services start, they will always be able to access storage.   Also allows them to be worked on independently.

app/config/storage.js
marctemp: Does this have knock-on effect as is, i.e. w/ those deployed in higher envs?

What is the remitting actions if so?

scripts/test
```diff
@@ -51,6 +51,9 @@ compose() {
 (
   cd "${projectRoot}"
 
+  # create test output folder
+  mkdir -p test-output
```
marctemp, line 55: Doesn't Sonarcloud use this for its metrics -- where they not running before or..?

marctemp (approved):
LGTM

app/config/storage.js
johnwatson484: It doesn't have any impact in higher environments.  These are just the default values used if none are provided through the environment variable.  For local development we don't set that env var so the defaults are used.

In higher envs, we're always deploying this service to Kubernetes through the Helm chart.  The Helm chart does set the env var through the config map and the value there was already the correct value of `statements`.

scripts/test
johnwatson484: The way the tests work is that when run they will create the `test-output` folder if it doesn't exist.  But because the tests are running in a docker container, the file is created as `root` user.  This actually prevents the container volume mapping from writing the test outputs to it.

The fix is to just run something like `sudo chmod 777 ./test-output` to change the permissions and re-run the tests again.  You're likely only ever going to see that error the first time you run the tests.

Alternatively that can be avoided by creating the folder in advance of running the containerised tests as they'll be created with an appropriate permission.  So seems a no brainer just to include it in the test script so it's created before the container starts and new devs who clone the repo don't have that pain on first run.

SonarCloud does need that folder and if you look in the Jenkins pipeline code you'll see that before it runs the tests, it does make sure the folder is created first -> https://github.com/DEFRA/ffc-jenkins-pipeline-library/blob/dc769b23423d710593a4ddc176f435f4403afd05/src/uk/gov/defra/ffc/Tests.groovy#L12

So the pipeline already does the equivalent of this.

PR #19, opened 2022-10-13
Remove reduction reasons

johnwatson484:
https://eaflood.atlassian.net/browse/PAY-2404


app/generator/content/part2/get-table/map-funding-rows.js
```diff
@@ -1,15 +1,16 @@
 const toCurrencyString = require('../../../to-currency-string')
-const mapReductionIds = require('./map-reduction-ids')
+const orderFunding = require('./order-funding')
```
johnwatson484, line 2: There was no ordering on funding previously, this caused the table rows to jump around on each generation.  Adding order so consistent and logical flow.

AbidemiAdio (commented):
in https://github.com/DEFRA/ffc-pay-statement-generator/blob/support-reductions/app/generator/content/part2/get-table/index.js 
consider changing ` getTable = (fundings, reductions)`  to  ` getTable = (fundings)`

AbidemiAdio (approved):
QED

PR #20, opened 2022-10-13
PAY-2391 Send data to CRM Topic

AbidemiAdio:


johnwatson484 (commented):
Added some comments.

Summary is we don't need to handle retries because it's already covered and we can get the blob path from the response to publishing it to storage, rather than it need to be passed in through config.

johnwatson484:
This line will return a blob upload response, which includes the full URL to storage.

https://github.com/DEFRA/ffc-pay-statement-generator/blob/55d439b22479083fcc909ce33d40c1950b58f56d/app/generator/publish.js#L21

docs/asyncapi.yaml
```diff
@@ -15,6 +15,34 @@ components:
...
+                type: number
+                description: Firm Reference Number
+                minimum: 1000000000
+                maximum: 9999999999
+                example: 1234567890
+              blobUrl:
+                type: string
+                description: url address of statement file that has been published unto blob
+                example: 'https://devffcpayst1001.blob.core.windows.net/statements/outbound/FFC_PaymentStatement_SFI_2022_1234567890_2022090615023001.pdf'
+          type: 'uk.gov.pay.statement.crm'
```
johnwatson484, line 44: Probably tweak the example to be `myBlobStorageAccount.blob.core` or something so it's not our actual account name.

test/unit/messaging/create-crm-message.test.js
```diff
@@ -0,0 +1,69 @@
...
+    })
+  }
+})
+
+jest.mock('../../../app/messaging/crm/crm-schema')
+const schema = require('../../../app/messaging/crm/crm-schema')
+
+const createCrmMessage = require('../../../app/messaging/crm/create-crm-message')
+const mockStatement = require('../../mocks/statement-data')
+const BLOB_URL = 'https://devffcpayst1001.blob.core.windows.net/statements/outbound/FFC_PaymentStatement_SFI_2022_1234567890_2022080515301012.pdf'
```
johnwatson484, line 19: Same comment, would amend the name

test/unit/messaging/send-crm-message.test.js
```diff
@@ -0,0 +1,40 @@
...
+    })
+  }
+})
+
+jest.mock('../../../app/messaging/crm/create-crm-message')
+const createCrmMessage = require('../../../app/messaging/crm/create-crm-message')
+
+const sendCrmMessage = require('../../../app/messaging/crm/send-crm-message')
+const mockStatement = require('../../mocks/statement-data')
+const BLOB_URL = 'https://devffcpayst1001.blob.core.windows.net/statements/outbound/FFC_PaymentStatement_SFI_2022_1234567890_2022080515301012.pdf'
```
johnwatson484, line 19: And again

app/messaging/crm/send-crm-message.js
```diff
@@ -0,0 +1,12 @@
+const { crmTopic } = require('../../config')
+const { MessageSender } = require('ffc-messaging')
+const createCrmMessage = require('./create-crm-message')
+
+const sendCrmMessage = async (statement, filename) => {
```
johnwatson484, line 5: Suggest renaming `filename` to `blobUrl` so it's clearer what it is.

johnwatson484 (commented):
Just minor comments.

app/messaging/crm/send-crm-message.js
AbidemiAdio: How did I not change this? copy and paste effect.

johnwatson484 (approved):
LGTM

johnwatson484 (approved):

PR #26, opened 2022-11-19
Add payment schedule

johnwatson484:


app/generator/content/statement/summary/get-address.js
```diff
@@ -0,0 +1,12 @@
+const { millimetresToPoints } = require('../../../conversion')
+
+const getAddress = (businessName, address) => {
+  let fullAddress = `${businessName}\n`
+  const addressLines = Object.values(address)
+  addressLines.filter(x => x !== undefined && x !== null && x !== '').forEach(x => {
```
marctemp, line 6: Are we convinced `address` value will be an array?

app/messaging/process-message.js
```diff
@@ -0,0 +1,22 @@
+const generateDocument = require('../generator')
+const util = require('util')
+const validateRequest = require('./validate')
+const { VALIDATION } = require('../errors')
+
+const processStatementMessage = async (message, receiver) => {
```
marctemp, line 6: `processMessage`

app/messaging/schemas/schedule.js
```diff
@@ -0,0 +1,12 @@
+const Joi = require('joi')
+const address = require('./address')
+const scheme = require('./scheme')
+
+module.exports = Joi.object({
+  businessName: Joi.string().required(),
+  frn: Joi.number().integer().min(1000000000).max(9999999999).required(),
+  sbi: Joi.number().integer().min(105000000).max(999999999).required(),
+  email: Joi.string().optional().allow('', null),
```
marctemp, line 9: Let's move these into their own schema definitions a la `address` and `scheme`

app/types.js
```diff
@@ -0,0 +1,4 @@
+module.exports = {
```
marctemp, line 1: Is `types` descriptive enough for this (also remind me what these are, I think they're just meta data for Azure events, right?)

test/unit/messaging/crm/create-crm-message.test.js
```diff
@@ -0,0 +1,104 @@
...
+    expect(result).toStrictEqual(crmMessage)
+  })
+
+  test('should throw Error when schema validate throws Error', async () => {
+    schema.validate.mockReturnValue({ error: 'Not a valid object' })
+    expect(() => createCrmMessage(mockStatement, FILENAME, STATEMENT)).toThrow(Error)
+  })
+})
+
+describe('send crm message for payment schedule', () => {
```
simonnineteen90, line 57: Nice use of an additional `describe` first time ive seen it 'out in the wild' since we discussed it!

simonnineteen90 (commented):
Few suggestions only

app/generator/get-definition.js
```diff
@@ -1,14 +1,14 @@
 const styles = require('./styles')
```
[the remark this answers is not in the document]
johnwatson484: The term "definition" is from the `PDFMake` docs.  However, it does refer to it as `docDefinition`.

https://pdfmake.github.io/docs/0.1/document-definition-object/styling/

What if we renamed it `get-document-definition` so it's clearer as to it's intent but still relatable to PDF make docs by glancing at the name?

app/messaging/schemas/adjustment.js
```diff
@@ -0,0 +1,7 @@
+const Joi = require('joi')
```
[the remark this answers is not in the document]
johnwatson484: I've added some for FRN and SBI.  The others are all required/optional and I don't think there's any value overriding the default Joi "x is required"

test/unit/messaging/crm/create-crm-message.test.js
johnwatson484: Thought I'd be bold!

app/generator/get-definition.js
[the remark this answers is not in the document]
simonnineteen90: Agree that would be better.

simonnineteen90 (approved):

PR #27, opened 2022-11-21
127816 Publish message should be updated with API link

AbidemiAdio:


app/config/index.js
```diff
@@ -5,12 +5,16 @@ const storageConfig = require('./storage')
 
 // Define config schema
 const schema = Joi.object({
-  env: Joi.string().valid('development', 'test', 'production').default('development')
+  env: Joi.string().valid('development', 'test', 'production').default('development'),
+  ffcApiPath: Joi.string().required(),
```
johnwatson484, line 9: I'd maybe suggest renaming this from `ffcApiPath` as it's not really clear what it's used for.  It's used later as the get route appended to the query route, so what about `getStatementPath` or something?

app/config/index.js
```diff
@@ -5,12 +5,16 @@ const storageConfig = require('./storage')
...
+  statementReceiverEndpoint: Joi.string().required()
 })
 
 // Build config
 const config = {
-  env: process.env.NODE_ENV
+  env: process.env.NODE_ENV,
+  ffcApiPath: process.env.FFC_API_PATH,
```
johnwatson484, line 16: Same comment as above about name of env var

app/generator/index.js
```diff
@@ -14,7 +14,7 @@ const generateStatement = async (statement) => {
   const pdfDoc = printer.createPdfKitDocument(docDefinition)
   const { filename, blobUrl } = await publish(pdfDoc, statement, moment(timestamp).format('YYYYMMDDHHmmssSS'))
   await sendPublishMessage(statement, filename)
-  await sendCrmMessage(statement, blobUrl)
+  await sendCrmMessage(statement, blobUrl, filename)
```
johnwatson484, line 17: Do we still need the `blobUrl` or can we remove it now?  If we're publishing this outside our infrastructure for CRM, then just including the one link in the message they need would probably simpler.

app/messaging/crm/crm-schema.js
```diff
@@ -3,5 +3,6 @@ const Joi = require('joi')
 module.exports = Joi.object({
   sbi: Joi.number().integer().min(105000000).max(999999999).required(),
   frn: Joi.number().integer().min(1000000000).max(9999999999).required(),
-  blobUrl: Joi.string().required()
+  blobUrl: Joi.string().required(),
+  apiLink: Joi.string().required()
```
johnwatson484, line 7: There is a `joi.string().uri()` valildation option that we could use

docker-compose.test.yaml
```diff
@@ -17,7 +17,8 @@ services:
       - ./package.json:/home/node/package.json
     environment:
       AZURE_STORAGE_CONNECTION_STRING: DefaultEndpointsProtocol=http;AccountName=devstoreaccount1;AccountKey=Eby8vdM02xNOcqFlqUwJPLlmEtlCDXJ1OUzFT50uSRZ6IFsuFq2UVErCz4I6tq/K1SZFPTOtr/KBHBeksoGMGw==;BlobEndpoint=http://ffc-pay-statement-azurite:10000/devstoreaccount1;
-
+      STATEMENT_RECEIVER_ENDPOINT: http://localhost:3000
```
johnwatson484, line 20: What's the value of setting this to localhost for the test definition?

Localhost will be local to the generator container only which will never resolve.  Would it be better to just mock endpoint in the tests so we don't get any side effect from this?

docker-compose.yaml
```diff
@@ -27,6 +27,8 @@ services:
       POSTGRES_USERNAME: ${POSTGRES_SCHEMA_USERNAME:-postgres}
       AZURE_STORAGE_ACCOUNT_NAME: ${AZURE_STORAGE_ACCOUNT_NAME:-devstoreaccount1}
       AZURE_STORAGE_USE_CONNECTION_STRING: "true"
+      FFC_API_PATH: ffcApi
+      STATEMENT_RECEIVER_ENDPOINT: http://ffc-pay-statement-receiver
```
johnwatson484, line 31: If this is intended to use the Kubernetes traffic then should be `http://ffc-pay-statement-receiver.ffc-pay-<ENVIRONMENT>`

But I don't think this should use the Kubernetes network because the requests will be coming from outside the cluster from CRM, so this value will result in an unusable link being created.

The traffic route will be `CRM -> App Gateway -> AKS -> NGINX -> ffc-pay-statement-receiver`.  So the link needs to be for a route we'll setup in App Gateway.

Not sure if you've added something in App Config to override these values.  We could maybe do something clever in the helm chart to build them depending on the `environmentCode`.

For path, same suggestion as above about naming.  GDS also has some standards we could factor in here -> https://www.gov.uk/guidance/gds-api-technical-and-data-standards

I'd also suggest we should include a version number too, eg `/v1/` so we can extend the API in future safely.

app/messaging/crm/create-crm-message.js
```diff
@@ -1,10 +1,12 @@
 const schema = require('./crm-schema')
+const { ffcApiPath, statementReceiverEndpoint } = require('../../config')
 
-const createCrmMessage = (statement, blobUrl) => {
+const createCrmMessage = (statement, blobUrl, filename) => {
   const crm = {
-    sbi: statement.sbi,
+    apiLink: `${statementReceiverEndpoint}/${ffcApiPath}/statement?filename=${filename}`,
```
johnwatson484, line 6: Only part of the path is used within the variable, `/statement` and the query string are outside.  Should we include all but the query string, or the query string too?

What about changing the API to be `/statementReceiverEndpoint/ffcApiPath/statement`  rather than needing a querystring?

helm/ffc-pay-statement-generator/values.yaml
```diff
@@ -39,6 +39,8 @@ container:
   azureStorageCreateContainers: false
   storageContainer: statements
   storageFolder: outbound
+  statementVersion: v1
+  statementReceiverEndpoint: http://ffc-pay-statement-receiver.ffc-pay
```
johnwatson484, line 43: This will need overwriting via a Value from App Config once we have DNS and App Gateway setup.

I can't find a ticket that covers that so can you make sure that we have one that includes:
- Setup App Gateway
- Setup DNS
- Update App Config with ingress values for `ingress.server` and `ingress.endpoint` for `ffc-pay-statement-receiver`
- Update App Config with corresponding values in this repository

johnwatson484 (approved):
Happy to approve this but just calling out the missing pieces of the puzzle we need captured on our board.

PR #28, opened 2022-11-30
Update CRM link

johnwatson484:
https://dev.azure.com/defragovuk/DEFRA-FFC/_workitems/edit/133856﻿

### DIFF
--- a/docker-compose.yaml
+++ b/docker-compose.yaml
@@ -28,7 +28,7 @@ services:
       AZURE_STORAGE_ACCOUNT_NAME: ${AZURE_STORAGE_ACCOUNT_NAME:-devstoreaccount1}
       AZURE_STORAGE_USE_CONNECTION_STRING: "true"
       STATEMENT_VERSION: v1
-      STATEMENT_RECEIVER_ENDPOINT: http://ffc-pay-statement-receiver:3000
+      STATEMENT_RECEIVER_ENDPOINT: http://localhost:3021
 
   ffc-pay-statement-generator-postgres:
     image: postgres:11.4-alpine
--- a/docker-compose.yaml
+++ b/docker-compose.yaml
@@ -11,7 +11,7 @@ services:
       - ffc-pay-statement-generator-postgres
     environment:
       APPINSIGHTS_CLOUDROLE: ffc-pay-statement-generator-local
-      APPINSIGHTS_INSTRUMENTATIONKEY: ${APPINSIGHTS_INSTRUMENTATIONKEY}      
+      APPINSIGHTS_INSTRUMENTATIONKEY: ${APPINSIGHTS_INSTRUMENTATIONKEY}
       MESSAGE_QUEUE_HOST: ${MESSAGE_QUEUE_HOST}
       MESSAGE_QUEUE_PASSWORD: ${MESSAGE_QUEUE_PASSWORD}
       MESSAGE_QUEUE_USER: ${MESSAGE_QUEUE_USER}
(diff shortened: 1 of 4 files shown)


helm/ffc-pay-statement-generator/values.yaml
```diff
@@ -40,7 +40,10 @@ container:
   storageContainer: statements
   storageFolder: outbound
   statementVersion: v1
-  statementReceiverEndpoint: http://ffc-pay-statement-receiver.ffc-pay
+  statementReceiverEndpoint: https://ffc-pay-statement-receiver
+
+ingress:
+  server: example.com
```
johnwatson484, line 46: Using `ingress` object will mean existing app config will populate this value in each environment without the need to manage additional duplicate configuration.

helm/ffc-pay-statement-generator/values.yaml
```diff
@@ -40,7 +40,10 @@ container:
   storageContainer: statements
   storageFolder: outbound
   statementVersion: v1
-  statementReceiverEndpoint: http://ffc-pay-statement-receiver.ffc-pay
+  statementReceiverEndpoint: https://ffc-pay-statement-receiver
```
johnwatson484, line 43: Original is the endpoint for the internal cluster network.  Requests will originate from CRM, not a Kubernetes pod so need to update to support external gateway.

This would first be `https` and also need to be dynamic as the subdomain changes in each environment.  The `ConfigMap` handles mapping the dynamic values.

Incidentally the current approach would not work beyond sandpit as all namespaces in higher environments have suffixes, eg `ffc-pay-dev` so even if this was a link for AKS services, would still need similar changes in the `ConfigMap`.

docker-compose.yaml
```diff
@@ -28,7 +28,7 @@ services:
       AZURE_STORAGE_ACCOUNT_NAME: ${AZURE_STORAGE_ACCOUNT_NAME:-devstoreaccount1}
       AZURE_STORAGE_USE_CONNECTION_STRING: "true"
       STATEMENT_VERSION: v1
-      STATEMENT_RECEIVER_ENDPOINT: http://localhost:3000
+      STATEMENT_RECEIVER_ENDPOINT: http://localhost:3021
```
johnwatson484, line 31: For local development, receiver is bound to localhost port `3021` so updating so link is usable locally.

helm/ffc-pay-statement-generator/templates/config-map.yaml
```diff
@@ -26,5 +26,5 @@ data:
   AZURE_STORAGE_CONTAINER: {{ quote .Values.container.storageContainer }}
   AZURE_STORAGE_FOLDER: {{ quote .Values.container.storageFolder }}
   STATEMENT_VERSION: {{ quote .Values.container.statementVersion }}
-  STATEMENT_RECEIVER_ENDPOINT: {{ quote .Values.container.statementReceiverEndpoint }}
+  STATEMENT_RECEIVER_ENDPOINT: {{ .Values.container.statementReceiverEndpoint }}-{{ .Values.environmentCode }}.{{ .Values.ingress.server }}
```
simonnineteen90, line 29: Does this value need to be quoted as it was before?

johnwatson484: No can't do that.  If we quoted that value then it would only quote the first template.  eg  {{ quote .Values.container.statementReceiverEndpoint }}-{{ quote  .Values.environmentCode }}.{{ quote .Values.ingress.server }}

would result in something like "https://my-endpoint"-"my-env"."my-server"

Which obvs isn't a real URL.  There may be a way to combine and quote through go templating, but we know this will work as is because we've done the same else where and also similar above in the Service Bus topics

simonnineteen90 (approved):

PR #33, opened 2023-02-03
Removed unnecessary line from part3.js

Kirsten1822:
Ticket link: https://dev.azure.com/defragovuk/DEFRA-FFC/_sprints/taskboard/DEFRA-FFC%20Payments%20Team/DEFRA-FFC/Payment%20Service/Pay%20Sprint%2048?workitem=142967


test/unit/generator/content/statement/summary/summary.test.js
```diff
@@ -34,31 +34,43 @@ describe('generate summary', () => {
...
 
   test('includes SBI', () => {
     const result = summary(mockStatement)
     expect(result.stack[5].columns[1].text).toBe(123456789)
   })
 
+  test('includes agreement number title', () => {
+    const result = summary(mockStatement)
+    expect(result.stack[6].columns[0].text).toBe('Agreement reference number:')
+    // expect(result.table.body[0][0].stack[4].columns[0].text).toBe('Agreement reference number:')
```
marctemp, line 48: Trailing comment

test/unit/generator/content/statement/summary/summary.test.js
```diff
@@ -34,31 +34,43 @@ describe('generate summary', () => {
...
+  })
+
+  test('includes agreement number value', () => {
+    const result = summary(mockStatement)
+    expect(result.stack[6].columns[1].text).toBe('SFI1234567')
+    // expect(result.table.body[0][0].stack[4].columns[1].text).toBe('SFI1234567')
```
marctemp, line 54: Trailing comment

docs/asyncapi.yaml
```diff
@@ -1,7 +1,7 @@
 asyncapi: 2.4.0
 info:
   title: FFC Payment Statement Generator
-  version: '1.0.0'
+  version: '1.11.1'
```
marctemp, line 4: This version doesn't correlate with the version of the service at large -- meant to version the document -- we don't actually have a process around this, if you had any ideas... 👀

docs/asyncapi.yaml
```diff
@@ -52,156 +51,148 @@ components:
...
       type: string
       description: The business name
-      example: 'FFC Ltd'
-    SBI:
-      type: number
-      description: Single Business Identifier
-      minimum: 105000000
-      maximum: 999999999
-      example: 123456789
+      example: 'Summer Days Farm'
```
marctemp, line 91: All example denoted as a string don't need quotes around it

docs/asyncapi.yaml
```diff
@@ -52,156 +51,148 @@ components:
...
-      - year
-      - frequency
-      - agreementNumber
+      - area
+      - annualValue
+      - quarterlyValue
+      - quarterlyReduction
+      - quarterlyPayment
       properties:
         name:
```
marctemp, line 145: These should all be schemas and referenced

docs/asyncapi.yaml
```diff
@@ -52,156 +51,148 @@ components:
...
+    SBI:
+      type: number
+      description: Single Business Identifier
+      minimum: 105000000
+      maximum: 999999999
+      example: 123456789
     Address:
       type: object
       properties:
         line1:
```
marctemp, line 111: These should all be schemas and referenced

docs/asyncapi.yaml
```diff
@@ -214,50 +205,100 @@ components:
...
+      - invoiceNumber
+      - reference
+      - dueDate
+      - settled
+      - calculated
       - value
+      - period
       properties:
-        reason:
+        invoiceNumber:
```
marctemp, line 248: These should all be schemas and referenced

README.md
```diff
@@ -60,6 +60,20 @@ Use Docker Compose to run service locally.
...
+Key = type
+```
+
+```
+Type = String
+```
+
+```
+Value: uk.gov.pay.statement
+```
```
marctemp, line 75: Can include all of the code snippets under 1 set of multiple line code block (```)

docs/asyncapi.yaml
Kirsten1822: There's some instances where numbers, for example, shorten when not in ' ', so in instances like that they might have to be kept in single quote marks

docs/asyncapi.yaml
Kirsten1822: What do you mean exactly? Wouldn't referencing 'Funding' be reference enough?

docs/asyncapi.yaml
Kirsten1822: What should all be schemas and referenced?

docs/asyncapi.yaml
Kirsten1822: What should all be schemas and referenced?

docs/asyncapi.yaml
```diff
@@ -214,50 +205,100 @@ components:
         name:
```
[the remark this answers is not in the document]
Kirsten1822: What should all be schemas and referenced?

docs/asyncapi.yaml
marctemp: So this `Funding` object has got a number of properties

If we convert all of those properties into schemas, i.e. name, level, area, etc. then reference them inside this `Funding` object schema.

There are a number of benefits:

1. We can easily see what fields/objects exist within any potential message -- rather than them being hidden inside another object like here
2. We can reuse them inside other objects, e.g. `Funding` and `FundingTotal` share a lot of same properties
3. Provides a single source of truth: define once, use lots

docs/asyncapi.yaml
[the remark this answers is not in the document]
marctemp: Comment above

docs/asyncapi.yaml
marctemp: See previous comment

docs/asyncapi.yaml
```diff
@@ -1,7 +1,7 @@
 asyncapi: 2.4.0
 info:
   title: FFC Payment Statement Generator
-  version: '1.0.0'
+  version: '1.0.1'
```
marctemp, line 4: Let's not quote and let async implicitly cast it to string where it does

docs/asyncapi.yaml
```diff
@@ -37,170 +37,113 @@ components:
...
-        scheme:
-          $ref: '#/components/schemas/Scheme'
-        payments:
+        funding:
           type: array
           items:
-            $ref: '#/components/schemas/Payment'
-        funding:
+            - $ref: '#/components/schemas/Funding'
+            - $ref: '#/components/schemas/FundingTotal'
```
marctemp, line 81: On Studio's sandbox, it states that there can be other items (which for this is true, but see if you can find an asyncapi format that represents this better, i.e. 1+ Fundings but only 1 FundingTotal)

Screenshot 2023-02-10 at 11 43 07

Here's an example of using `oneof`: https://github.com/asyncapi/generator/issues/427

marctemp: Also, if you could look at it stating that here can be other items for the message itself which shouldn't be the case as we should be validating against that

Screenshot 2023-02-10 at 11 50 28

docs/asyncapi.yaml
```diff
@@ -260,4 +187,166 @@ components:
...
+          $ref: '#/components/schemas/FundingQuarterlyReduction'
+        quarterlyPayment:
+          $ref: '#/components/schemas/FundingQuarterlyPayment'
+    FundingName:
+      type: string
+      description: The funding option name
+      example: Arable and horticultural soils
+    FundingTotalName:
+      type: string
+      description: The funding option name
```
marctemp, line 223: Description should not be the same as `FundingName` above, "The funding total name"

docs/asyncapi.yaml
```diff
@@ -260,4 +187,166 @@ components:
...
+      example: Total
+    FundingLevel:
+      type: string
+      description: The funding option level
+      example: Introductory
+    FundingTotalLevel:
+      type: string
+      description: The funding option level
```
marctemp, line 231: Description should not be the same as FundingLevel above, "The funding total level"

docs/asyncapi.yaml
```diff
@@ -1,7 +1,7 @@
 asyncapi: 2.4.0
 info:
   title: FFC Payment Statement Generator
-  version: '1.0.0'
+  version: '1.0.1'
   description: Generate PDF payment statements
 channels:
   payment/request:
```
marctemp, line 7: As per our AsyncAPI standard doc, we want to use the Topic/Subscription as the name of the channel for this: https://eaflood.atlassian.net/wiki/spaces/SFI/pages/4177723458/AsyncAPI+Structure#Channels

docs/asyncapi.yaml
```diff
@@ -37,170 +37,113 @@ components:
...
-                type: url
                 description: api link for calling file stream of statement file that has been published unto blob
                 example: 'https://myStatementRetreivalApiEndpoint/statement-receiver/statement/v1/FFC_PaymentStatement_SFI_2022_1234567890_2022090615023001.pdf'  
               documentType:
                 type: string
                 enum:
                   - Payment statement
                 description: document type
-                example: 'Payment statement'
+                example: Payment statement
```
marctemp, line 47: This looks incomplete, could you add the complete enum list here (I think it's just something like payment schedule or similar)

docs/asyncapi.yaml
Kirsten1822: The best I can do in regards to the initial comment, with my current knowledge, is using a combination on anyOf and oneOf. I got the following information from searching around online:

oneOf – validates the value against exactly one of the subschemas
allOf – validates the value against all the subschemas
anyOf – validates the value against any (one or more) of the subschemas

Kirsten1822: Though, in trying the above, it seems like it interferes with the example message which is not what we would want

docs/asyncapi.yaml
```diff
@@ -1,10 +1,10 @@
 asyncapi: 2.4.0
 info:
   title: FFC Payment Statement Generator
-  version: '1.0.0'
+  version: 1.0.1
   description: Generate PDF payment statements
 channels:
-  payment/request:
+  ffc-pay-statements/ffc-pay-statement-generator:
```
marctemp, line 7: Wouldn't include the Subscription name because as per our naming conventions, it should be the name of the service :)

docs/asyncapi.yaml
```diff
@@ -19,9 +19,11 @@ components:
...
       contentType: application/json
       payload:
+        additionalProperties: false
         type: object
         properties:
           body:
+            additionalProperties: false
             type: object
             properties:
               sbi:
```
marctemp, line 29: For SBI and FRN coming in, I believe these are to be string -- dbl check that though

docs/asyncapi.yaml
```diff
@@ -37,171 +39,120 @@ components:
...
       required:
       - businessName
-      - sbi
       - frn
+      - sbi
       - address
-      - scheme
-      - payments
       - funding
+      - fundingTotal
```
marctemp, line 70: This property isn't in the object

docs/asyncapi.yaml
```diff
@@ -37,171 +39,120 @@ components:
...
                 description: document type
-                example: 'Payment statement'
+                example: 
+                  items:
+                    - Payment statement
+                    - Payment schedule
               scheme:
                 type: string
                 enum:
                   - SFI
```
marctemp, line 57: Again, seems incomplete

docs/asyncapi.yaml
marctemp: Nice, okay good research. Let's add these also to the Confluence page!

docs/asyncapi.yaml
```diff
@@ -37,171 +39,120 @@ components:
...
-      description: The business email address
-      example: farm@farms.com
+    SBI:
+      type: number
+      description: Single Business Identifier
+      minimum: 105000000
+      maximum: 999999999
+      example: 123456789
     Address:
       type: object
```
marctemp, line 116: Which of these properties are required?

marctemp: Can it have additional properties?

docs/asyncapi.yaml
```diff
@@ -37,171 +39,120 @@ components:
...
+      description: The fifth line of the address
+      example: Some County
+    Postcode:
+      type: string
+      description: The postcode
+      example: FA1 2PY
     Funding:
+      additionalProperties: false
       type: object
       required:
```
marctemp, line 157: Level and Rate are marked as not required, just dbl checking this is correct?

I believe level isn't due to Moorlands stuff, maybe rate comes into tht too?

docs/asyncapi.yaml
```diff
@@ -260,4 +195,167 @@ components:
         value:
           type: string
           description: The reduction value
-          example: 75.00
+          example: '75.00'
+    FundingTotal:
+      additionalProperties: false
+      type: object
+      required:
```
marctemp, line 202: Level and Rate not required?

docs/asyncapi.yaml
marctemp: Or, whatcha think on this?

I think Confluence page says to use both (suppose to help us devs quickly see which Topic/Sub to send msg to).?

Kirsten1822: Yeah the Confluence page says to use Topic Name/Subscription Name - I also think you suggested in one of your previous comments that the channel name should follow said naming convention on the Confluence page [https://eaflood.atlassian.net/wiki/spaces/SFI/pages/4177723458/AsyncAPI+Structure#Useful-Tips]

docs/asyncapi.yaml
Kirsten1822: Should I just remove this reference to it?

docs/asyncapi.yaml
Kirsten1822: Not sure what you mean here - could you elaborate? What do you believe to be missing?

docs/asyncapi.yaml
Kirsten1822: I'm not sure, would need a second opinion

docs/asyncapi.yaml
Kirsten1822: Again, not sure. That's just how it was labelled before I started working on the ticket

docs/asyncapi.yaml
Kirsten1822: https://github.com/DEFRA/ffc-pay-statement-generator/pull/33#discussion_r1104334837

docs/asyncapi.yaml
Kirsten1822: Well the format for the example message I used to generate a statement seemed to use a number for SBI and FRN, so I assume not a string. As this is CRM though, I can't imagine the SBI and FRN numbers being classed as a string here when the SBI and FRN numbers in Payment Statement are classed as numbers

docs/asyncapi.yaml
marctemp: Sounds good -- let's enforce it

docs/asyncapi.yaml
marctemp: Ah sorry, thought I left this under PaymentRequest's defintion

Just dbl check what is coming in -- it comes from `ffc-pay-statement-constructor` and I think a requirement was to send everything downstream as string

docs/asyncapi.yaml
marctemp: This suggests there's only 1 valid value for `Scheme` being the `enum` value `SFI`

Let's get a complete list of the possible schemes we support

docs/asyncapi.yaml
marctemp: Yup

docs/asyncapi.yaml
marctemp: Can't have additional values.

I _think_ the only required address lines are `Postcode` -- again, dbl check -- `ffc-pay-statement-constructor` might help

docs/asyncapi.yaml
marctemp: Yeah, unfort'ly this `async` doc hasn't been created 100% accurately.

Acc'ly looking at the schemas in this application code it would suggest that v few fields are required -- hmmmm

docs/asyncapi.yaml
Kirsten1822: In that case I will leave as is :)

docs/asyncapi.yaml
Kirsten1822: I checked out the tests in ffc-pay-statement-constructor and it seemed that the mock FRN and SBI weren't stored as strings, so I guess it's okay for them to be numbers?

docs/asyncapi.yaml
Kirsten1822: Where can I find the possible schemes support? Is it somewhere in ffc-pay-statement-generator?

docs/asyncapi.yaml
Kirsten1822: I checked the template given on this ticket and it seems that the required address lines are town, county and postcode to be displayed

docs/asyncapi.yaml
Kirsten1822: I'm going to put both down as required just in case

Kirsten1822: I found out that level, area and rate are actually all optional, so I'm going to change those now

docs/asyncapi.yaml
Kirsten1822: I'm gonna ask John to see if he knows :)

docs/asyncapi.yaml
Kirsten1822: I double checked in the code and both are numbers

docs/asyncapi.yaml
Kirsten1822: John said SFI is the only one supported atm, so I'm going to leave this as it is

marctemp (approved):
All improvements made -- great job with the asyncapi, let's get the rest of the services' docs as good as this!

PR #40, opened 2023-05-12
Save down document reference

marctemp:
https://dev.azure.com/defragovuk/DEFRA-FFC/_workitems/edit/181436


AbidemiAdio (approved):
Ticket scope well covered, but in general, we need to restructure files/folders grouping for better navigation.

PR #44, opened 2023-06-09
177639-Schedule content update for top up

AbidemiAdio:


app/generator/content/schedule/part3/get-table.js
```diff
@@ -10,27 +12,29 @@ const getTable = (schedule) => {
...
+          { text: toCurrencyString(x.value), style: 'tableNumber' }
+        ]))
       ]
     }
   }
 
   return table
 }
 
+const getRowSpan = (paymentType, rowsCount, order) => {
```
marctemp, line 34: Let's move into own file and add unit tests around it

app/generator/content/schedule/part1/index.js
```diff
@@ -11,11 +11,10 @@ const part1 = (schedule) => {
     stack: [
       { image: `${imagePath}/rpa-logo.jfif`, fit: [millimetresToPoints(200), millimetresToPoints(25)], style: 'scheduleLogo' },
       getAddress(schedule.businessName, schedule.address),
-      { text: `${schedule.scheme.name}`, style: 'header1' },
-      { text: 'Revised payment schedule', style: 'subTitle' },
-      getBusinessName(schedule.businessName),
       getSBI(schedule.sbi),
```
marctemp, line 14: This displays wrong title should be `SBI:` but is `Single business identifier (SBI):`

Also, we have 2 of the same content for both statement (`app/generator/content/statement/summary/get-sbi.js`) and schedule (`app/generator/content/schedule/part1/get-sbi.js`), can we move this into the common area like `app/generator/content/get-business-name.js` and reference it in both cases.

app/generator/content/schedule/part1/index.js
```diff
@@ -11,11 +11,10 @@ const part1 = (schedule) => {
...
-      getAgreementReference(schedule.scheme.agreementNumber)
+      getBusinessName(schedule.businessName),
+      getAgreementReference(schedule.scheme.agreementNumber),
```
marctemp, line 16: This displays wrong title should be `Agreement number:` but is `Agreement reference number:`

Also, we have 2 of the same content for both statement (`app/generator/content/statement/part1/get-agreement-number.js`) and schedule (`app/generator/content/schedule/part1/get-agreement-reference.js`), can we move this into the common area like `app/generator/content/get-business-name.js` and reference it in both cases.

app/generator/content/schedule/part3/get-table.js
```diff
@@ -1,6 +1,8 @@
 const toCurrencyString = require('../../../../generator/to-currency-string')
```
marctemp, line 1: Need to rely whether the grey background for the table and payment summary is correct as template does not have any background to it.

app/generator/content/schedule/part3/get-table.js
```diff
@@ -10,27 +12,29 @@ const getTable = (schedule) => {
...
-          ...schedule.map(x => ({ text: toCurrencyString(x.value), style: 'tableNumber' }))
-        ],
-        [
           { text: 'Payment period', style: 'tableHeader' },
-          ...schedule.map(x => ({ text: x.period, style: 'tableNumber' }))
-        ]
+          { text: 'Amount', style: 'tableHeader' }
+        ],
+        ...schedule.map(x => ([
+          { rowSpan: getRowSpan(x.paymentType, rowsCount, x.order), text: x.paymentType, style: 'tableNumber' },
```
marctemp, line 23: Need to see whether happy with `Quarterly payment` or whether they want it plural.

app/generator/content/schedule/part3/get-table.js
```diff
@@ -10,27 +12,29 @@ const getTable = (schedule) => {
...
+          { text: x.period, style: 'tableNumber' },
```
marctemp, line 24: Payment periods are different to template.

`March 2023` to your `Mar 2023`
`February 2023 to April 2023` to your `Feb-Apr 2023`

Need to verify this is expected.

test/unit/generator/content/schedule/part3/get-table.test.js
```diff
@@ -8,27 +8,27 @@ describe('get table', () => {
     mockSchedule = JSON.parse(JSON.stringify(require('../../../../../mocks/mock-schedule').topUpSchedule))
   })
 
-  test('includes table with three rows', () => {
+  test('includes table with number of colums equal schedule length + 1 ', () => {
```
marctemp, line 11: number of rows* equal to the schedule length + 1

app/generator/content/schedule/part3/get-table.js
```diff
@@ -10,27 +12,29 @@ const getTable = (schedule) => {
     fillColor: '#d9d9d9',
```
marctemp, line 12: Can you add missing test checking the fill value of the table

test/mocks/objects/schedule.js
```diff
@@ -1,37 +1,31 @@
...
+    order: 2,
+    dueDate: '01/03/2023',
+    paymentType: QUARTERLY,
+    period: 'Dec-Feb 2023',
+    value: '325.00'
+  },
+  {
+    order: 3,
+    dueDate: '01/06/2023',
+    period: 'Mar-May 2023',
```
marctemp, line 21: Can you flip these 2 keys around so its common ww/ above, i.e. `paymentType` then  `period`.

Same w/ below.

test/unit/generator/content/schedule/part1/index.test.js
```diff
@@ -20,11 +20,11 @@ describe('schedule Part 1', () => {
```
marctemp, line 20: I think this test and all `index` tests should be under integration seeing as its result is from calling multiple units which we've tested individually.

What do you think?

marctemp: Missing checks for the outputs of `getSBI`, `getBusinessName` and `getAgreementReference`

OOS for this ticket but can you add a ticket to come back to add these.

test/unit/generator/content/schedule/part1/index.test.js
```diff
@@ -20,11 +20,11 @@ describe('schedule Part 1', () => {
...
+    expect(result.stack[5].text).toBe(`${topUpSchedule.scheme.name} (${topUpSchedule.scheme.shortName}) revised payment schedule `)
   })
 
-  test('includes subtitle', () => {
+  test('header style is set to header3', () => {
     const result = part1(topUpSchedule)
-    expect(result.stack[3].text).toBe('Revised payment schedule')
+    expect(result.stack[5].style).toBe('header3')
   })
 })
```
marctemp, line 30: Can you add a test that ensures `unbreakable` is set to `true`

test/integration/schedule/part2/index.test.js
```diff
@@ -6,7 +6,7 @@ describe('schedule part 2', () => {
   describe('Get top-up description', () => {
```
marctemp, line 6: Can you update the folder/file name to match the path under the app folder

marctemp: Would like your & Anwar's/Edi's opinion on this, how should we be testing this for this file.

We have fantastic unit tests for the 2 potential funcs it calls `getTopUpDescription` and `getReductionDescription`. But the tests in this are an inferior copy-paste from these meaning:
1. it's not testing anything new
2. we need to change tests in 2 places if the output of those funcs change

I guess what we're wanting to test is that the right func is called in the correct scenario and the output is as we're expecting.

Just questioning the best way of doing that.

AbidemiAdio: Your conclusion that the content of the two test files are the same is wrong. End of first paragraph of  `getTopUpDescription` test reads increased while that of `getReductionDescription` reads decreased.

The standard practice is to create test file for each function file. Are you suggesting we keep the functions together?
If your answer is yes, then we are deviating from the original plan of separating Top-up format implementation from that of reduction.  We agreed on separating them so that the two format are not tightly coupled, hence, implementation of any required change in one of the format can easily be achieved without much impact on the other.

marctemp: No, not suggesting moving the funcs together.

marctemp: Sorry, didn't mean to suggest content of each func is the same, Ik they're not.

Saying that we have unit tests checking the content for each func. This func is deciding which func to use and returning that. So the tests for this file should just be on the decision aspect not the content itself because that's covered in the lower unit tests.

app/generator/content/schedule/part1/index.js
AbidemiAdio: `Agreement reference number` now changed to `Agreement number` in schedule, therefore content of the 2 files are no more the same.

app/generator/content/schedule/part3/get-table.js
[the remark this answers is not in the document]
AbidemiAdio: Sample of generated schedule was presented to the team on 12/06/2023, everyone is happy with the alignments.

AbidemiAdio: Background has been removed.

app/generator/content/schedule/part3/get-table.js
AbidemiAdio: Sample of generated schedule was presented to the team on 12/06/2023, everyone is happy with the content.

app/generator/content/schedule/part3/get-table.js
AbidemiAdio: Sample of generated schedule was presented to the team on 12/06/2023, everyone agreed that it should be presented as it is for now. If further consultation requires change to be made, new ticket would be raised to that effect.

app/generator/content/schedule/part3/get-table.js
AbidemiAdio: Sample of generated schedule was presented to the team on 12/06/2023, it was agreed that there should be no filled value, so, it has been removed, hence no test is required.

app/generator/content/schedule/part3/get-help-info.js
```diff
@@ -1,29 +1,41 @@
 const getHelpInfo = () => {
```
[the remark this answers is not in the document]
AbidemiAdio: Unit test added.

test/unit/generator/content/schedule/part1/index.test.js
AbidemiAdio: Checks for outputs are in the index test under integration folder.

test/unit/generator/content/schedule/part1/index.test.js
AbidemiAdio: Code updated

test/integration/schedule/part2/index.test.js
AbidemiAdio: It has been resolved, contents are not the same.

app/generator/content/schedule/part1/index.js
AbidemiAdio: Sorry I didn't add update, the files and expected result are not the same, hence, better to keep them apart as they are.

simonnineteen90 (approved):

PR #48, opened 2023-06-21
Remove singleton constraint

marctemp:
https://dev.azure.com/defragovuk/DEFRA-FFC/_workitems/edit/181446


helm/ffc-pay-statement-generator/values.yaml
```diff
@@ -19,7 +19,7 @@ containerSecret:
 
 deployment:
   minReplicas: 1
-  maxReplicas: 1
+  maxReplicas: 8
```
AbidemiAdio, line 22: Do we need up to 8?

marctemp: Tbh these numbers are v much pulled out of thin air w/ some guess-timates thrown in:
1. I made this the largest due to its simple msg-in, msg-out processing therefore there is no wasted processing waiting around for all data sources (Constructor), it doesn't do 2 separate processing jobs and/or rely on a 3rd party service (Publisher) so its the most suited for multi-processing
2. Due to this, I envisioned this some-what similar in processing to Enrichment which is set at upto 16 replicas.

These numbers will need re-evaluated w/ proper performance testing in which we may increase the replica numbers and decrease its memory/cpu requirements or vice-versa.

Happy to bump down to 4.

AbidemiAdio (approved):
QED

PR #49, opened 2023-06-22
194874 - Logo update in all existing documents

AbidemiAdio:


simonnineteen90 (commented):
Code looks good to me - two question / suggestions
1. Can the old logo be removed from the repo if its no longer being used on the statement?
2. Have you considered the brand guidelines that accompany the logo (downloaded from ticket) it suggest there should be more space above the logo - top margin.

AbidemiAdio:
Added top margin to logo and removed old logo file

PR #50, opened 2023-06-26
Sending email to team for schedules

marctemp:
https://dev.azure.com/defragovuk/DEFRA-FFC/_workitems/edit/197629

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -1,3 +1,3 @@
-@Library('defra-library@v-9') _
+@Library('defra-library@temp-remove-npm-audit') _
 
 buildNodeJs()
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "ffc-pay-statement-generator",
-  "version": "2.4.1",
+  "version": "2.4.2",
   "lockfileVersion": 2,
   "requires": true,
   "packages": {
     "": {
       "name": "ffc-pay-statement-generator",
-      "version": "2.4.1",
+      "version": "2.4.2",
       "license": "OGL-UK-3.0",
       "dependencies": {
         "@azure/identity": "2.0.5",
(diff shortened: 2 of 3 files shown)


app/messaging/process-message.js
```diff
@@ -20,6 +20,4 @@ const processMessage = async (message, receiver) => {
   }
 }
 
-module.exports = {
-  processMessage
-}
+module.exports = processMessage
```
marctemp, line 23: Functions being exported from their file should not be wrapped in `{}` this this should be done in `index.js` module files when exporting multiple functions from within.

helm/ffc-pay-statement-generator/templates/config-map.yaml
```diff
@@ -21,10 +21,12 @@ data:
   AZURE_STORAGE_ACCOUNT_NAME: {{ quote .Values.container.azureStorageAccountName }}
   AZURE_STORAGE_CONNECTION_STRING: {{ quote .Values.container.azureStorageConnectionString }}
   AZURE_STORAGE_USE_CONNECTION_STRING: {{ quote .Values.container.azureStorageUseConnectionString }}
-  AZURE_STORAGE_USE_CONNECTION_STRING: {{ quote .Values.container.azureStorageUseConnectionString }}
```
marctemp, line 24: Duplicate to above

AbidemiAdio:
I thought we agreed during refinement that all notify implementation will take place in the service statement-publisher., hence, sending notification email to staff will be implemented in statement-publisher.

marctemp:
> I thought we agreed during refinement that all notify implementation will take place in the service statement-publisher., hence, sending notification email to staff will be implemented in statement-publisher.

That was the original thought.

However, we currently don't send schedules downstream to Publisher: https://github.com/DEFRA/ffc-pay-statement-generator/blob/main/app/config/index.js#L11 & https://dev.azure.com/defragovuk/DEFRA-FFC/_git/DEFRA-FFC-PLATFORM?path=/services/ffc-pay/appConfig/appConfig.prd.json&version=GBmain&line=519&lineEnd=520&lineStartColumn=1&lineEndColumn=1&lineStyle=plain&_a=contents

This is because the Publisher service publishes whatever it receives to the email provided in the request which we cannot do  with schedules because CRM needs them to be combined with other documents and sent from their end.

Therefore, it would have been a very large change to the way the Publisher service works for this workaround.

Hence, I added it here.

AbidemiAdio:
I will review statement-publisher to see if there is simple logic we can apply. The essence of creating different services is to separate responsibilities, bringing email-dispatch into generator looks like we are bringing something out-of-scope into the service.

AbidemiAdio (approved):

PR #51, opened 2023-07-03
194720 Reduction Schedule content update for 0 payment

AbidemiAdio:


app/generator/content/schedule/part2/get-reduction-zero-description.js
```diff
@@ -0,0 +1,11 @@
+const getReductionZeroDescription = (schedule) => {
```
marctemp, line 1: Just double checking this name is right.

Can we get a zero payment from a top-up rather than reduction?

E.g. PR1 is 1000 | Paid 250 | PR2 is 200 | PR3 is 250

Would this follow the same business path?

app/generator/content/schedule/part2/get-recovery-description.js
```diff
@@ -0,0 +1,17 @@
+const getRecoveryDescription = (schedule) => {
```
marctemp, line 1: Ticket only concerns 0 reductions.

Can you split out the recovery functionality into a separate PR

app/generator/content/schedule/part2/get-reduction-zero-description.js
AbidemiAdio: The name is right, this ticket only covers path of reduction to zero from positive. The example you gave is reduction from negative (recovery) to zero, we would need to consider some AR entries to treat this scenario. Issues around getting AR data was raised during our last meeting, it was concluded that the scenario of recovery to zero is possible but very rare, hence we should treat case as reduction to zero for now.

app/generator/content/schedule/part2/get-recovery-description.js
AbidemiAdio: Yes, the ticket only concern 0 reductions. Treating zero reduction involves breaking standard reduction scenario into three possible scenarios (standard-reduction, reduction-to-zero and recovery) , hence the tickets take care of the three possible scenarios, but only implement changes to reduction-to-zero format. There is a ticket for implementing recovery format and it is coming as another PR.

app/generator/content/schedule/part2/index.js
```diff
@@ -1,8 +1,18 @@
...
 const part2 = (schedule) => {
-  return schedule.adjustment.adjustmentValue > 0 ? getTopUpDescription(schedule) : getReductionDescription(schedule)
+  const remainingAmount = schedule.remainingAmount
+  switch (true) {
+    case (remainingAmount > 0):
+      return schedule.adjustment.adjustmentValue > 0 ? getTopUpDescription(schedule) : getReductionDescription(schedule)
+    case (remainingAmount === 0):
+      return getReductionZeroDescription(schedule)
+    case (remainingAmount < 0):
+      return getRecoveryDescription(schedule)
```
marctemp, line 14: Do we have existing functionality for recovery?

Why do we need to check and implement recovery functionality for this ticket?

Can we just leave this PR to handle the 1st (as-is) and 2nd (ticket ask) case functionality and leave this recovery aspect for the recovery ticket?

Like, we're adding dummy functionality for recovery when we have a sep ticket to target and focus on it with the signed off version.

From a testing POV because this exists, we need to test it which means QA needs to test this but it doesn't fit under the ticket's ask and expands its scope.

test/mocks/components/remaining-amount.js
```diff
@@ -1 +1,11 @@
-module.exports = 200.01
+const topUpRemainingAmount = 200.01
+const reductionRemainingAmount = 200.01
+const reductionZeroRemainingAmount = 0
```
marctemp, line 3: `0.00` for explicit 'we expect this to be a float' (I think it'll get cast to an int anyway)

test/unit/messaging/schemas/remaining-amount.test.js
```diff
@@ -1,7 +1,6 @@
-const REMAINING_AMOUNT = require('../../../mocks/components/remaining-amount')
```
marctemp, line 1: Now we need to test out this schema against each of these valid test cases.

The describes on lines 39 and 65 need replicated for when remaining amount is top-up, when remaining amount is zero value, when remaining amount is... etc

OOS for this ticket but let's add a ticket in the backlog.

marctemp: Can shorten to 1 line

`const { topUpRemainingAmount: REMAINING_AMOUNT } = require('../../../mocks/components/remaining-amount')`

test/unit/messaging/schemas/remaining-amount.test.js
```diff
@@ -1,7 +1,6 @@
...
-
+const { topUpRemainingAmount } = require('../../../mocks/components/remaining-amount')
+const REMAINING_AMOUNT = topUpRemainingAmount
 const schema = require('../../../../app/messaging/schemas/remaining-amount')
-
```
marctemp, line 4: I think the whitespace aids readability between imports and variable value that will change between test cases

test/mocks/components/remaining-amount.js
```diff
@@ -1 +1,11 @@
-module.exports = 200.01
+const topUpRemainingAmount = 200.01
```
marctemp, line 1: These values aren't just thought up to match 'must be a float' and for recovery 'must be negative'.

These are meant to collate to real test scenarios given in our int tests, by providing top-up from `mock-schedule` results in a remaining amount of this, whereas providing reduction from `mock-schedule` results in a remaining amount of line 2 because the rest of the data and the processing of the system output tht.

This is a smell that valuable integration tests haven't been done.

test/unit/generator/content/schedule/part2/index.test.js
```diff
@@ -4,9 +4,15 @@ const getTopUpDescription = require('../../../../../../app/generator/content/sch
 jest.mock('../../../../../../app/generator/content/schedule/part2/get-reduction-description')
```
marctemp, line 4: Can we better structure this test file?

We have 4 (should only be 3 as-is) scenarios to test for:
1.  top-up
2. reduction
3. zero value reduction
4. (shouldn't be) recovery]

For each of these we pass in their respective schedule and expect 2 things:
1. calls its expected function
2. doesn't call any of the others

Let's group this file into tht format

describe('top-up')
calls func
doesn't call funcA
doesn't call funcB
etc

describe('reduction')
calls func
doesn't call funcA
doesn't call funcB
etc

marctemp: Also, the functions in this aren't complete.

For each of ensuring we call top-up function we aren't checking:
1. That its called once
2. What its called with

Please add

marctemp: Finally, we don't have test cases when schedule being passed in is not any of these, i.e. undefined, null, {}, etc

app/generator/content/schedule/part3/get-summary.js
```diff
@@ -0,0 +1,18 @@
...
+const getSummary = (schedule) => {
+  const remainingAmount = schedule.remainingAmount
+  switch (true) {
+    case (remainingAmount > 0):
+      return schedule.adjustment.adjustmentValue > 0 ? getTopUpSummary(schedule.adjustment) : getReductionSummary(schedule)
+    case (remainingAmount === 0):
+      return getReductionZeroSummary(schedule)
+    case (remainingAmount < 0):
+      return getRecoverySummary(schedule)
+  }
```
marctemp, line 15: No case when remainingAmount is undefined because schedule doesn't have a remainingAmount key

test/unit/generator/content/schedule/part2/index.test.js
marctemp: Same for part3

test/unit/generator/content/schedule/part3/get-summary.test.js
```diff
@@ -0,0 +1,51 @@
...
+  test('Ensure top-up-schedule does not return reduction summary', () => {
+    getSummary(topUpSchedule)
+    expect(getReductionSummary).not.toHaveBeenCalled()
+  })
+
+  test('Ensure reduction-schedule does not returns top-up summary', () => {
+    getSummary(reductionSchedule)
+    expect(getTopUpSummary).not.toHaveBeenCalled()
+  })
+})
```
marctemp, line 51: See comments for part2 index.js unit test

test/unit/generator/content/schedule/part3/index.test.js
```diff
@@ -7,8 +7,14 @@ const getReductionSummary = require('../../../../../../app/generator/content/sch
 jest.mock('../../../../../../app/generator/content/schedule/part3/get-top-up-summary')
```
marctemp, line 7: We have changed this functionality and need to change the tests for it.

This function no longer directly calls `getTopUpSummary`, `getRecoverySummary` but rather just `getSummary`.

The functionality in this should be moved to getSummary unit test

app/generator/content/schedule/part2/index.js
AbidemiAdio: Removed all instances of recovery.

test/mocks/components/remaining-amount.js
AbidemiAdio: integer 0 is a member of numeric type, hence 0 value is valid, for better readability, I have changed to 0.00

test/unit/messaging/schemas/remaining-amount.test.js
AbidemiAdio: The are is already a ticket for reviewing remainingAmount test files. I am sure we would be able to address auto test with Joi components and their usage.

test/mocks/components/remaining-amount.js
AbidemiAdio: There is separation of responsibility between statement-constructor and statement-generator. statement-constructor deals with calculation and correctness of data manipulation, while statement-generator deal with validating and checking data-value from statement-constructor, then generates documents based of data-values that are received from statement-constructor.

test/unit/generator/content/schedule/part2/index.test.js
AbidemiAdio: The code block used is strict switch (switch(true)), hence any condition not listed in switch block would totally be ignored.

app/generator/content/schedule/part3/get-summary.js
AbidemiAdio: remainingAmount can never be undefined, it only allows numeric, it does not allow any. There can only be 3 possible values of numeric   >0, <0 and =0, hence all possible scenario are taken care of. any other condition or logic flow will be redundant.

test/unit/generator/content/schedule/part2/index.test.js
```diff
@@ -4,32 +4,122 @@ const getTopUpDescription = require('../../../../../../app/generator/content/sch
...
+
 const part2 = require('../../../../../../app/generator/content/schedule/part2')
 
-const { topUpSchedule, reductionSchedule } = require('../../../../../mocks/mock-schedule')
+const { topUpSchedule, reductionSchedule, reductionZeroSchedule, negativeRemainingAmountSchedule } = require('../../../../../mocks/mock-schedule')
 
 describe('schedule part 2', () => {
   afterEach(() => {
     jest.clearAllMocks()
   })
```
marctemp, line 17: Whitespace missing

test/unit/generator/content/schedule/part2/index.test.js
```diff
@@ -4,32 +4,122 @@ const getTopUpDescription = require('../../../../../../app/generator/content/sch
...
+      part2(topUpSchedule)
+      expect(getTopUpDescription).toHaveBeenCalled()
+    })
+
+    test('Ensure getTopUpDescription is called once', () => {
+      part2(topUpSchedule)
+      expect(getTopUpDescription).toHaveBeenCalledTimes(1)
+    })
+
+    test('Ensure getTopUpDescription is called with topUpSchedule.adjustment', () => {
```
marctemp, line 29: with topUpSchedule not topUpSchedule.adjustment

test/unit/generator/content/schedule/part2/index.test.js
```diff
@@ -4,32 +4,122 @@ const getTopUpDescription = require('../../../../../../app/generator/content/sch
...
+    test('Ensure getReductionpDescription is not called', () => {
+      part2(reductionZeroSchedule)
+      expect(getReductionDescription).not.toHaveBeenCalled()
+    })
   })
 
-  test('Ensure reduction-schedule does not returns top-up description', () => {
-    part2(reductionSchedule)
-    expect(getTopUpDescription).not.toHaveBeenCalled()
+  describe('description with schedule of negativeRemainingAmount', () => {
```
marctemp, line 99: Can you standardise the names of these describes in this file

test/unit/generator/content/schedule/part3/index.test.js
```diff
@@ -88,34 +55,56 @@ describe('schedule part 3', () => {
       expect(getTable).toHaveBeenCalledTimes(1)
     })
 
-    test('Ensure getTable is called with reductionSchedule.schedule', () => {
+    test('Ensure getTable is called with reductionSchedule', () => {
```
marctemp, line 58: with reductionSchedule.schedule not with reductionSchedule

test/unit/generator/content/schedule/part3/get-summary.test.js
```diff
@@ -0,0 +1,125 @@
...
+
+const getSummary = require('../../../../../../app/generator/content/schedule/part3/get-summary')
+
+const { topUpSchedule, reductionSchedule, reductionZeroSchedule, negativeRemainingAmountSchedule } = require('../../../../../mocks/mock-schedule')
+
+describe('get summary', () => {
+  afterEach(() => {
+    jest.clearAllMocks()
+  })
+  describe('top-up summary', () => {
```
marctemp, line 18: Missing whitesapce

test/unit/generator/content/schedule/part3/get-summary.test.js
```diff
@@ -0,0 +1,125 @@
...
+      expect(getTopUpSummary).not.toHaveBeenCalled()
+    })
+
+    test('Ensure getReductionpSummary is not called', () => {
+      getSummary(reductionZeroSchedule)
+      expect(getReductionSummary).not.toHaveBeenCalled()
+    })
+  })
+
+  describe('summary with schedule of negativeRemainingAmount', () => {
```
marctemp, line 99: Can we standardise the describes names in this file

test/unit/generator/content/schedule/part3/get-summary.test.js
```diff
@@ -0,0 +1,125 @@
...
+      getSummary(negativeRemainingAmountSchedule)
+      expect(getReductionSummary).toHaveBeenCalled()
+    })
+
+    test('Ensure getReductionSummary is called once', () => {
+      getSummary(negativeRemainingAmountSchedule)
+      expect(getReductionSummary).toHaveBeenCalledTimes(1)
+    })
+
+    test('Ensure getReductionSummary is called with reductionSchedule', () => {
```
marctemp, line 110: with negativeRemainingAmountSchedule not with reductionSchedule

test/mocks/mock-schedule.js
```diff
@@ -26,16 +23,32 @@ const baseSchedule = {
 
 const topUpSchedule = {
   ...baseSchedule,
+  remainingAmount: topUpRemainingAmount,
```
marctemp, line 26: Does a top-up have a remainingAmount?

test/integration/generator/content/schedule/part3/index.test.js
```diff
@@ -0,0 +1,67 @@
+jest.mock('../../../../../../app/generator/content/schedule/part3/get-table')
+const { getTable } = require('../../../../../../app/generator/content/schedule/part3/get-table')
+
+const part3 = require('../../../../../../app/generator/content/schedule/part3')
+const { topUpSchedule, reductionSchedule, reductionZeroSchedule } = require('../../../../../mocks/mock-schedule')
```
marctemp, line 5: Missing whitespace

test/integration/generator/content/schedule/part3/index.test.js
```diff
@@ -0,0 +1,67 @@
...
+
+describe('schedule part 3 integration with getSummary components', () => {
+  afterEach(() => {
+    jest.clearAllMocks()
+  })
+
+  test('Ensure getTable is called once', () => {
```
marctemp, line 12: Remmove test.

This is a unit test not integration.
This is duplicate from here: https://github.com/DEFRA/ffc-pay-statement-generator/pull/51/files#diff-ba0b49ed636c32320433e3ce38d8eafa14b83020b08918d89a4803b366bfbc4dR16
For each scenario not just top-up

test/integration/generator/content/schedule/part3/index.test.js
```diff
@@ -0,0 +1,67 @@
...
+      const result = part3(topUpSchedule)
+      expect(result.stack[0].style).toBe('header3')
+    })
+
+    test('Ensure top-up summary title text is "Payment schedule"', () => {
+      const result = part3(topUpSchedule)
+      expect(result.stack[0].text).toBe('Payment schedule')
+    })
+
+    test('includes unbreakable instruction when top up schedule', () => {
```
marctemp, line 28: Do not need, "when top up schedule" in test name because the describe already tells us this is part of the top-up schedule tests

marctemp: Subsequent for other tests under describes

test/integration/generator/content/schedule/part3/index.test.js
```diff
@@ -0,0 +1,67 @@
...
+      const result = part3(topUpSchedule)
+      expect(result.unbreakable).toBeTruthy()
+    })
+  })
+
+  describe('reduction zero schedule part 3', () => {
```
marctemp, line 34: Wrong describe name, should be "reduction schedule part 3"

test/integration/generator/content/schedule/part3/index.test.js
```diff
@@ -0,0 +1,67 @@
...
+      const result = part3(reductionZeroSchedule)
+      expect(result.stack[0].text).toBe('Payment schedule')
+    })
+
+    test('includes unbreakable instruction when reduction schedule', () => {
+      const result = part3(reductionZeroSchedule)
+      expect(result.unbreakable).toBeTruthy()
+    })
+  })
+})
```
marctemp, line 67: OOS for this ticket because it's been neglected but we should have integration tests that check what the result of calling these index functions return, i.e. the large JSON objects we get back and interrogate them based on what we want to know, i.e. they are stitched together properly (making sure not to repeat our unit tests)

These large JSONs should not be too difficult to make up our expected return due to having the mock components to build from.

test/unit/generator/content/schedule/part3/get-reduction-zero-summary.test.js
```diff
@@ -0,0 +1,41 @@
+const toCurrencyString = require('../../../../../../app/generator/to-currency-string')
+const getReductionZeroSummary = require('../../../../../../app/generator/content/schedule/part3/get-reduction-zero-summary')
+const { reductionZeroSchedule } = require('../../../../../mocks/mock-schedule')
```
marctemp, line 3: Missing whitespace

test/mocks/components/remaining-amount.js
marctemp: Yeee understand that -- I just want our mock test values to mimic as close as possible a real top-up schedule or reduction-schedule, etc

Which because we just pick up messages from a Topic isn't difficult to get.

test/unit/generator/content/schedule/part2/index.test.js
AbidemiAdio: No white space missing, please check latest file : 
https://github.com/DEFRA/ffc-pay-statement-generator/blob/194721-Reduction-Schedule-content-update-for-0-payment/test/unit/generator/content/schedule/part2/index.test.js

test/integration/generator/content/schedule/part3/index-with-get-summary-components.test.js
```diff
@@ -0,0 +1,84 @@
+jest.mock('../../../../../../app/generator/content/schedule/part3/get-table')
+const { getTable } = require('../../../../../../app/generator/content/schedule/part3/get-table')
+
+const part3 = require('../../../../../../app/generator/content/schedule/part3')
+
+const { topUpSchedule, reductionSchedule, reductionZeroSchedule, negativeRemainingAmountSchedule } = require('../../../../../mocks/mock-schedule')
+
+describe('schedule part 3 integration with getSummary components', () => {
```
marctemp, line 8: Test file name doesn't refer to a file.

Should be schedule/part3/index -- consistent w/ other modules and their tests

This describe name can just be part 3

marctemp: I would also argue these tests are unit not integration because we are mocking a func call w/in and don't care what these return.

marctemp (approved):
All suggestions actioned

PR #53, opened 2023-07-13
136721 - Generate payment schedule with a recovery

AbidemiAdio:


test/mocks/mock-schedule.js
```diff
@@ -42,16 +42,16 @@ const reductionZeroSchedule = {
...
 }
 
-const negativeRemainingAmountSchedule = {
+const recoverySchedule = {
   ...baseSchedule,
-  remainingAmount: negativeRemainingAmount,
-  schedule: negativeRemainingAmountScheduleTimeline,
+  remainingAmount: recoveryRemainingAmount,
+  schedule: recoveryScheduleTimeline,
   adjustment: reductionAdjustment
```
marctemp, line 49: Do we need a recovery adjustment obj?

GodsonLeigh (changes requested):
A couple of comments to look at regarding the use of reduction in a place where we have switched to recovery. Other than that all good.

samplackett (approved):
approved following previous reviews from Leigh and only change to version number since

PR #63, opened 2023-08-28
207629-Updating statement layout changes

AbidemiAdio:


app/generator/content/get-help-info.js
```diff
@@ -1,7 +1,7 @@
-const getHelpInfo = () => {
+const getHelpInfo = (partLabel) => {
```
GodsonLeigh, line 1: Is partLabel the best name for this or could this be something more descriptive to the function? Would using moreInformation be more understandable.

app/generator/content/statement/part3/get-table/index.js
```diff
@@ -15,12 +15,12 @@ const getTable = (fundings) => {
         [
           { text: 'Standard', style: 'tableHeader' },
           { text: 'Level', style: 'tableHeader' },
-          { text: 'Rate', style: 'tableHeader' },
+          { text: 'Rate (£)', style: 'tableHeader' },
```
GodsonLeigh, line 18: Does this need the additional text in the ticket including as it was never clarified - rate per ha is included on the layout we are following, but not included in the code.

GodsonLeigh (commented):
Only two minor comments, one regarding the naming and another for a text value that was never fully clarified. Other than that, it works well and the refactor cleans it all up nicely.

GodsonLeigh (changes requested):
Main request is clarification on Rate per Ha being omitted.

GodsonLeigh (approved):
Looks good. Thanks

PR #70, opened 2023-10-24
237025 create different content template of each document type for each scheme sfi 22 and sfi 23

AbidemiAdio:


GodsonLeigh (approved):
Looks great to me.

PR #85, opened 2024-01-16
changed asyncAPI pattern, added readme, updated unit tests for new fo…

GodsonLeigh:
following on for the publisher asyncAPI pattern changes, the generator service needed modifying to supply the correct pattern


docs/asyncapi.yaml
```diff
@@ -470,14 +470,18 @@ components:
...
+      description: The PDF filename for the payment schedule, formatted as product code using 3 to 6 uppercase characters, descriptor in PascalCas
+      example: FFC_PaymentSchedule_SFI_2022_1234567890_202305181018.pdf
+      pattern: ^[A-Z]{3,6}\_[A-Z]([A-Z0-9]*[a-z][a-z0-9]*[A-Z]|[a-z0-9]*[A-Z][A-Z0-9]*[a-z])[A-Za-z0-9]*\_[A-Z]{3,6}_\d{4}_\d{10}_\d{12}\.pdf$
     FilenameStatement:
       type: string
-      description: The PDF filename for the payment statement
-      example: FFC_PaymentStatement_SFI_2022_1234567890_2023051810180298.pdf
+      description: The PDF filename for the payment statement, formatted as product code using 3 to 6 uppercase characters, descriptor in PascalCa
+      example: FFC_PaymentStatement_SFI_2022_1234567890_202305181018.pdf
+      pattern: ^[A-Z]{3,6}\_[A-Z]([A-Z0-9]*[a-z][a-z0-9]*[A-Z]|[a-z0-9]*[A-Z][A-Z0-9]*[a-z])[A-Za-z0-9]*\_[A-Z]{3,6}_\d{4}_\d{10}_\d{12}\.pdf$
```
AbidemiAdio, line 483: Any reason for reducing the length of the time-stamp section of filename from 16 to 12?

GodsonLeigh: The spec file we received showed that the final element was to be 12 digits.

AbidemiAdio (dismissed):
QED

test/unit/generator/create-filename.test.js
```diff
@@ -124,14 +127,12 @@ describe('create filename', () => {
...
-    expect(result).toContain('MyScheme')
+    const result = getFilename(mockSchedule, timestamp, STATEMENT)
+    expect(result).toBe('FFC_PaymentStatement_SFI_2022_1234567890_2022080515301012.pdf')
   })
 
   test('removes spaces if schedule', () => {
-    mockSchedule.scheme.shortName = 'My Scheme'
     const result = getFilename(mockSchedule, timestamp, SCHEDULE)
-    expect(result).toContain('MyScheme')
+    expect(result).toBe('FFC_PaymentSchedule_SFI_2022_1234567890_2022080515301012.pdf')
```
AbidemiAdio, line 136: There should be  tests to prove the following:

1. No error message when filename composition is valid.
2. Expected filename is created when filename composition is valid
3. There is error message when filename composition is not valid.
4. The expected filename is not created when filename is not valid.

GodsonLeigh: I have added a test that checks that an incorrect filename creation results in the throw Error as well as the examples above that are checking that the final filename creation without spaces is in the correct format.

test/unit/generator/validate-filename.test.js
```diff
@@ -0,0 +1,45 @@
+const moment = require('moment')
+const getFilename = require('../../../app/generator/create-filename')
+const { STATEMENT } = require('../../../app/constants/document-types')
+
+let timestamp
+let errorStatement
```
AbidemiAdio, line 6: use mockStatement

test/mocks/error-statement.js
```diff
@@ -0,0 +1,92 @@
...
+    name: 'Total',
+    level: '',
+    area: '42.00',
+    rate: '',
+    annualValue: '1318.60',
+    quarterlyValue: '329.65',
+    quarterlyReduction: '125.00',
+    quarterlyPayment: '242.15'
+  }]
+}
```
AbidemiAdio, line 92: Could you please remove this file and use mockStatement?

test/unit/generator/validate-filename.test.js
GodsonLeigh: removed errorStatement and switched to mockStatement

test/mocks/error-statement.js
GodsonLeigh: removed errorStatement and switched to mockStatement

GodsonLeigh: deleted

AbidemiAdio (dismissed):
QED

JungaBunga (approved):

